PR #4, opened 2018-02-06
Converts the capturing service to the foreground

muthenberg:
This commit adds a notification that is shown while data capturing is
active. This prevents the system from killing the service, since it runs
as a foreground service.


datacapturing/src/main/res/values/strings.xml
```diff
@@ -1,4 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <string name="capturing_service_description">Use to collect sensor data (GPS, accelerometer, gyroscope, magnetometer) and send it to your local community, where it is used to improve the mobility infrastructure.</string>
+    <string name="capturing_active">Capturing ongoing</string>
```
hb0: "Capturing ongoing" klingt falsch. "Capturing in progress"

PR #5, opened 2018-02-12
Story/#mov 73 data sync

muthenberg:
Mal wieder ein etwas größerer Pull Request. Das meiste sollte allerdings Dokumentation sein. An Code ist eigentlich fast nur der SyncAdapter (und alles was man so dazu braucht) dazu gekommen.


datacapturing/src/main/AndroidManifest.xml
```diff
@@ -3,6 +3,10 @@
 
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
+    <uses-permission android:name="android.permission.WRITE_SYNC_SETTINGS" />
```
hb0: I personally prefer to annotate the requested permissions with the reason, why they were added. This helps when refactoring / cleaning up the code later on.

datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
```diff
@@ -30,13 +37,25 @@
...
  */
 public final class DataCapturingService {
 
+    private static final String TAG = "de.cyface.capturing";
+    private final static String ACCOUNT = "default_account";
+    private final static String ACCOUNT_TYPE = "de.cyface";
+    private final static String AUTHORITY = BuildConfig.provider;
+    public static final long SECONDS_PER_MINUTE = 60L;
+    public static final long SYNC_INTERVAL_IN_MINUTES = 60L;
+    public static final long SYNC_INTERVAL = SYNC_INTERVAL_IN_MINUTES * SECONDS_PER_MINUTE;
```
hb0: Any reason why you choose 60 minutes as sync interval? You might wanna write it down here ...

datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
```diff
@@ -191,9 +206,46 @@ public void disconnect() {
...
+        if (context.get() == null) {
+            throw new IllegalStateException("No valid context to enable data synchronization!");
+        }
+
+        Account account = getAccount();
+
+        boolean cyfaceAccountSyncIsEnabled = ContentResolver.getSyncAutomatically(account, AUTHORITY);
+        boolean masterAccountSyncIsEnabled = ContentResolver.getMasterSyncAutomatically();
+
+        if (cyfaceAccountSyncIsEnabled && masterAccountSyncIsEnabled) {
```
hb0: If one of both accounts is not enabled, we do nothing, not even logging this event down?  
Can/should this happen anyway?

datacapturing/src/main/res/values/strings.xml
```diff
@@ -1,8 +1,13 @@
...
+    <string name="app_name">Cyface</string>
     <string name="capturing_service_description">Used to collect sensor data (GPS, accelerometer, gyroscope, magnetometer) and to send it to your local community, where it is used to improve the mobility infrastructure.</string>
     <string name="capturing_active">Capturing in progress</string>
     <string name="cancel_capturing">Cancel Capturing</string>
     <string name="notification_title">Cyface</string>
     <string name="ticker_text">Ticker</string>
+
+    <string name="missing_data_error">Captured data was missing from message!</string>
+    <string name="unknown_message_error">Received unknown message %d from data capturing service!</string>
+    <string name="unable_to_rebind_error">"Unable to rebind. Context was null."</string>
```
hb0: Are the quotes inside the brackets >"..."< there on purpose?

persistence/src/main/java/de/cyface/persistence/MeasurementTable.java
```diff
@@ -19,65 +17,59 @@
...
      * Logging tag for Android logging.
-     * </p>
      */
     static final String TAG = "MeasurementTable";
     /**
-     * <p>
-     * This column is not used for anything, just to be able to create a measurement in the db without defining any values,
-     * see: http://stackoverflow.com/a/2663620
-     * </p>
+     * A boolean value which is either <code>true</code> (or 1 in SQLLite) if this measurement has been completed or
```
hb0: Looks much cleaner this way :+1:

synchronization/src/main/java/de/cyface/synchronization/SyncPerformer.java
```diff
@@ -0,0 +1,125 @@
...
+import java.net.ProtocolException;
+import java.net.URL;
+
+/**
+ * Performs the actual synchronisation with a provided server, by uploading meta data and a file containing
+ * measurements.
+ * 
+ * @author Klemens Muthmann
+ * @version 1.0.0
+ * @since 2.0.0
```
hb0: Version < Since? Does this make sense? If not, you might wanna check if this typo occurs in this PR multiple times ...

synchronization/src/main/java/de/cyface/synchronization/SyncPerformer.java
```diff
@@ -0,0 +1,125 @@
...
+
+            String userIdPart = addPart("userId", deviceIdentifier, boundary);
+            String measurementIdPart = addPart("measurementId", Long.valueOf(measurementIdentifier).toString(),
+                    boundary);
+
+            String fileHeader1 = "--" + boundary + "\r\n"
+                    + "Content-Disposition: form-data; name=\"fileToUpload\"; filename=\"" + fileName + "\"\r\n"
+                    + "Content-Type: application/octet-stream\r\n" + "Content-Transfer-Encoding: binary\r\n";
+
+            // FIXME This will only work correctly as long as we are using a ByteArrayInputStream. For other streams it
```
hb0: (Not sure if this is related:) I had troubles using ByteBuffer in the beginning as it's pointer is not always on the first entry ... In the end I figured it out, reading the documentation, see e.g. backend2/GpsPointSerializer, but I can't remember if the problem only was the mixture of read and write access on the Buffer without using .flip().

datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
muthenberg: Comes from the example code. There is no requirement on how to set this. So I just used that value.

datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
muthenberg: These are not accounts but system settings, which the user uses to allow or disallow an App or all Apps to communicate with the internet. If it is not allowed we are out of luck an there really is nothing we can do.

synchronization/src/main/java/de/cyface/synchronization/SyncPerformer.java
muthenberg: The version of the Framework is 2.0.0. (Our old app is considered 1.0.0). This call was created in version 2.0.0 of the framework so @since is 2.0.0. The version of the class is still 1.0.0 since it was created here. @since and @version have nothing to do with each other.

synchronization/src/main/java/de/cyface/synchronization/SyncPerformer.java
muthenberg: Works so far. There are test cases and as long as they do not fail I have no choice but assuming this code is ok. There will be a few extended tests prior to the campaign, that should uncover any error conditions.

PR #12, opened 2018-03-02
Adds data zipping based on Android Deflater

muthenberg:


synchronization/src/mock/java/de/cyface/synchronization/SyncPerformer.java
```diff
@@ -7,17 +7,44 @@
 import android.util.Log;
 
 /**
- * Created by muthmann on 26.02.18.
+ * A sync performer that does absolutely nothing and just prints its status to the terminal.
```
hb0: So what is it there for?

PR #14, opened 2018-03-04
Fix/#mov 192 deregister ui tracks

muthenberg:


synchronization/src/main/java/de/cyface/synchronization/CyfaceSyncAdapter.java
```diff
@@ -89,19 +95,28 @@ public void onPerformSync(final @NonNull Account account, final @NonNull Bundle
         MeasurementSerializer serializer = new MeasurementSerializer();
         SyncPerformer syncer = new SyncPerformer(getContext());
         SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(getContext());
+        Cursor syncableMeasurementsCursor = null;
+        AccountManager accountManager = AccountManager.get(getContext());
+        // Account[] accounts AccountManager.get(getContext()).getAccountsByType(StubAuthenticator.ACCOUNT_TYPE);
```
hb0: unused code => can be removed ?

PR #22, opened 2018-04-16
Fix/#mov 227 android 8 sync stop

muthenberg:


build.gradle
```diff
@@ -37,7 +37,7 @@ ext {
     // Migrate to new startForeground method and test the DataCapturing extensively before
     // targeting version 26! See details here:
     // https://stackoverflow.com/questions/43251528/android-o-old-start-foreground-service-still-working
-    targetSdkVersion = 25
```
hb0: I guess you did follow the new guidelines on how to start a foreground service in API 26+? In this case you can drop the comment above this line.

datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingListener.java
```diff
@@ -9,7 +9,7 @@
  * {@link DataCapturingService#start(DataCapturingListener,de.cyface.datacapturing.model.Vehicle)}.
  *
  * @author Klemens Muthmann
- * @version 1.1.0
+ * @version 1.2.0
```
hb0: Why did the code version change but nothing else?

muthenberg: Because I reverted all changes in a following commit by hand.

build.gradle
muthenberg: Jepp. Seems to work good.

PR #30, opened 2018-05-11
Adds threadPool implementation for persisting Point3D data #MOV-252

hb0:


datacapturing/src/main/java/de/cyface/datacapturing/backend/CapturingProcess.java
```diff
@@ -165,7 +166,11 @@ public void onLocationChanged(final Location location) {
 
             synchronized (this) {
                 for (CapturingProcessListener listener : this.listener) {
-                    listener.onLocationCaptured(new GeoLocation(latitude, longitude, gpsTime, speed, gpsAccuracy));
+                    try {
+                        listener.onLocationCaptured(new GeoLocation(latitude, longitude, gpsTime, speed, gpsAccuracy));
+                    } catch (DataCapturingException e) {
+                        throw new IllegalStateException(e);
```
hb0: IllegalStateException is thrown when location data is captured without GpsFix (i.e. timestamp = 0L), that should be legit, right?

datacapturing/src/main/java/de/cyface/datacapturing/persistence/CapturedDataWriter.java
```diff
@@ -0,0 +1,153 @@
...
+                }
+                for (int startIndex = 0; startIndex < values.length; startIndex += MAX_SIMULTANEOUS_OPERATIONS) {
+                    int endIndex = Math.min(values.length, startIndex + MAX_SIMULTANEOUS_OPERATIONS);
+                    // BulkInsert is about 80 times faster than insertBatch
+                    client.bulkInsert(MeasuringPointsContentProvider.MAGNETIC_VALUE_POINTS_URI,
+                            Arrays.copyOfRange(values, startIndex, endIndex));
+                }
+
+            } catch (DataCapturingException e) {
+                // TODO: not sure if this is ok according to our SDK API?
```
hb0: If client == null it should be legit to throw an IllegalStateException, can this actually happen when used properly?

datacapturing/src/main/java/de/cyface/datacapturing/persistence/MeasurementPersistence.java
```diff
@@ -197,7 +113,13 @@ public void storeData(final @NonNull CapturedData data, final long measurementId
      * @param location The geo location to store.
      * @param measurementIdentifier The identifier of the measurement to store the data to.
      */
-    public void storeLocation(final @NonNull GeoLocation location, final long measurementIdentifier) {
+    public void storeLocation(final @NonNull GeoLocation location, final long measurementIdentifier)
+            throws DataCapturingException {
+        if (location.getTimestamp() == 0L) {
+            throw new DataCapturingException(
+                    "GeoLocations captured without GpsFix. Make sure you only capture data when a fix is available.");
```
hb0: This came from the 1.0.0 code so I migrated it back here as this should still be the case (timestamp 0L if no GpsFix) and such location data should never be stored, right?

datacapturing/src/main/java/de/cyface/datacapturing/backend/CapturingProcess.java
muthenberg: It is legit for framework functions, which you cannot change to throw a DataCapturingException. For all other cases use a typed exception.

PR #40, opened 2018-08-13
Story/#cy 3573 sdk in app integrieren

hb0:
Changes made during pair programming and commit #5989d0e1ce342d115db08a13392c04b626164d9e ontop.


muthenberg:
BTW.: Did you cancel the build? Shouldn't it run, before we merge this?

hb0:
As I wrote before: the build takes more than 600s which leads to an auto-cancel on your Bitrise account. I'll set it up again tomorrow on my account when I check the build PR-trigger bug u reported

hb0:
@muthenberg About the NotificationChannel description, see https://developer.android.com/training/notify-user/channels: For each different type of notification one should create a channel so that the user can decide demselves if they want to see the "Capturing ongoing" notifications, the "Warning" notifications, the "Info" Notifications, etc.

In order for the user to decide this, we should provide a description for each channel we create.
You are right - the stop button is currently not implemented so this part of the description is a bit ahead of it's time.

hb0:
Anmerkung: Vor dem Approven bitte kurz den merge commit prüfen, ob da alles korrekt gelaufen ist:
https://github.com/cyface-de/android-backend/pull/40/commits/c1caf78594f6ee228f296e1fb594abdc91e410a6

synchronization/src/cyface/java/de/cyface/synchronization/CyfaceSyncAdapter.java
```diff
@@ -37,20 +40,28 @@
 import de.cyface.persistence.GpsPointsTable;
 import de.cyface.persistence.MagneticValuePointTable;
 import de.cyface.persistence.MeasurementTable;
+import de.cyface.persistence.MeasuringPointsContentProvider;
 import de.cyface.persistence.RotationPointTable;
 import de.cyface.persistence.SamplePointTable;
 
 public final class CyfaceSyncAdapter extends AbstractThreadedSyncAdapter {
 
+    public final static String NOTIFICATION_CHANNEL_ID_WARNING = "cyface_warnings";
```
hb0: alt

PR #50, opened 2018-08-23
Task/#cy 3733 error handling rebased

hb0:
Adjusts the SDK to send specific Error Intent Broadcasts which can be identified and shown by the Cyface App.

Rebased to the top of dev => All tests run through

PR #52, opened 2018-08-23
Task/#cy 3974 move error handler to sdk

hb0:
Implementation of Feedback from PR review https://github.com/cyface-de/android-backend/pull/50 as far as reasonable. I created another task for refacoring the SyncProgress Listener to use the ErrorHandler.


datacapturing/src/main/java/de/cyface/datacapturing/StartUpFinishedHandler.java
```diff
@@ -39,7 +39,7 @@
      *
      * @param measurementIdentifier The identifier of the measurement that is captured by the started capturing process.
      */
```
hb0: This API changed recently. Implementing the new method resulted into a warning from the IDE that simple types (long) cannot have a NonNull flag. Thus, I removed the flag.

synchronization/src/cyface/java/de/cyface/synchronization/CyfaceSyncAdapter.java
```diff
@@ -110,19 +131,15 @@ public CyfaceSyncAdapter(final @NonNull Context context, final boolean autoIniti
     public void onPerformSync(final @NonNull Account account, Bundle extras, String authority,
             final @NonNull ContentProviderClient provider, final @NonNull SyncResult syncResult) {
         final Context context = getContext();
-
         Log.d(TAG, "Sync started.");
+
         final SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(context);
         final String deviceIdentifier = preferences.getString(SyncService.DEVICE_IDENTIFIER_KEY, null);
```
hb0: As this is only in the Cyface flavor I now let the sdk crash if it's used falsly.

synchronization/src/cyface/java/de/cyface/synchronization/CyfaceSyncAdapter.java
```diff
@@ -141,21 +158,15 @@ public void onPerformSync(final @NonNull Account account, Bundle extras, String
```
hb0: The do {} while {} content is definitelly too complex, I just sliced it a bit into prepare(), fill() and post() but we have to refactor this eventually.

hb0:
Anmerkung: Der Test ist im CI nur gefailt, weil das @Ignore flag nur im dev gefixt wurde. Passt also alles.

PR #58, opened 2018-09-04
Remove exception and correct string format

muthenberg:
This commit removes the exception thrown if a speed smaller than 0 is
captured. This obviously can happen on certain devices.

Secondly the commit fixes the wrong format string used in the error
reporting in the GeoLocations class. This should now not cause an
exception anymore.

### DIFF
--- a/datacapturing/src/main/java/de/cyface/datacapturing/model/GeoLocation.java
+++ b/datacapturing/src/main/java/de/cyface/datacapturing/model/GeoLocation.java
@@ -5,6 +5,7 @@ import java.util.Locale;
 import android.os.Parcel;
 import android.os.Parcelable;
 import android.support.annotation.NonNull;
+import android.util.Log;
 
 import de.cyface.datacapturing.DataCapturingService;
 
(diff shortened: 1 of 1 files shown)


datacapturing/src/main/java/de/cyface/datacapturing/model/GeoLocation.java
```diff
@@ -5,17 +5,22 @@
...
  * @author Klemens Muthmann
- * @version 1.2.0
+ * @version 1.2.1
  * @since 1.0.0
  */
 public class GeoLocation implements Parcelable {
+    /**
+     * The tag used to identify Logcat output from objects of this class.
+     */
+    private static final String TAG = "de.cyface.capturing";
```
hb0: The generic "capturing" TAG should be stored centrally not duplicated into each capturing class.

muthenberg: Correct. However I would like to avoid such ground breaking changes in the current master. We should do that as part of the integration and release for next years version

hb0: Sure ;-)

PR #63, opened 2018-09-11
Improvement/#cy 4076 clean up connection listeners

hb0:
Wegen der Deletes etwas größerer PR.

Moves the ConnectionListener interface to the SDK and merges duplicate Listener implementations and error handling.


datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
```diff
@@ -16,6 +16,7 @@
 
 import android.Manifest;
 import android.accounts.Account;
+import android.content.BroadcastReceiver;
```
muthenberg: Useless Import?

synchronization/src/cyface/java/de/cyface/synchronization/ConnectionListener.java
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * This event is called when the sync progress changed.
+     *
+     * @param transmittedPoints The number of points transmitted so far.
+     * @param pointsToTransmit The number of points to transmit.
+     * @param measurementId The measurement id of the measurement which is currently transmitted.
+     */
+    void onProgress(final long transmittedPoints, final long pointsToTransmit, final long measurementId);
+
+    void onSyncFinished();
```
muthenberg: documentation?

synchronization/src/cyface/java/de/cyface/synchronization/CyfaceSyncAdapter.java
```diff
@@ -114,7 +116,7 @@ private CyfaceSyncAdapter(final @NonNull Context context, final boolean autoInit
         this.rotationsUploadBatchSize = rotationsUploadBatchSize;
         this.directionsUploadBatchSize = directionsUploadBatchSize;
         progressListener = new HashSet<>();
-        addSyncProgressListener(new CyfaceSyncProgressListener(context));
+        addSyncProgressListener(new CyfaceConnectionListener(context));
```
muthenberg: Since this is not called SyncProgressListener anymore, this method should get a new name as well.

synchronization/src/cyface/java/de/cyface/synchronization/CyfaceSyncAdapter.java
```diff
@@ -217,26 +235,48 @@ private void syncMeasurement(final String authority, final ContentProviderClient
...
                 // from the Movebis flavor and the file uploader which we'll implement before releasing this.
                 // TODO: We should probably remove the unused isSynced flag from the points after
-                // we implemented the Cyface Byte Format synchronization. #
+                // we implemented the Cyface Byte Format synchronization. #CY-3592
 
                 // We delete the data of each point type separately to avoid #CY-3859 parcel size error.
                 deletePointsOfType(provider, authority, measurementSlice, geoLocationJsonMapper);
                 deletePointsOfType(provider, authority, measurementSlice, accelerationJsonMapper);
                 deletePointsOfType(provider, authority, measurementSlice, rotationJsonMapper);
                 deletePointsOfType(provider, authority, measurementSlice, directionJsonMapper);
```
muthenberg: Thoase 4 things should happen in loader.cleanMeasurement. There are two implementations of this method. The one for Movebis just deletes accelerations, rotations and directions, while the one for Cyface deletes the measurement.

synchronization/src/main/java/de/cyface/synchronization/ConnectionBroadcastReceiver.java
```diff
@@ -0,0 +1,84 @@
...
+            case SYNC_PROGRESS:
+                pointsToTransmit = intent.getLongExtra(SYNC_POINTS_TO_TRANSMIT, -1L);
+                final long transmittedPoints = intent.getLongExtra(SYNC_POINTS_TRANSMITTED, -1L);
+                final long measurementId = intent.getLongExtra(SYNC_MEASUREMENT_ID, -1L);
+                Validate.isTrue(pointsToTransmit > 0L);
+                Validate.isTrue(transmittedPoints > 0L);
+                Validate.isTrue(measurementId > 0L);
+
+                for (final ConnectionListener listener : connectionListener) {
+                    if (listener != null) {
```
muthenberg: Why is there a null check here, but not in the iterations above? Is it even possible for a listener to become null? Either way; either this check is useless or similar checks are missing for SYNC_FINISHED and SYNC_STARTED.

synchronization/src/cyface/java/de/cyface/synchronization/CyfaceSyncAdapter.java
hb0: The reason why we separately delete points here is:
- We still transmit measurements in slices, i.e. we want to delete the points after each successful slice transmission. This way it's no problem if the transmission is interrupted after a few slices.
- Only when a measurement was fully transmitted, we use the loader.cleanMeasurement method to drop everything related with this measurement (the measurement entry itself)

synchronization/src/main/java/de/cyface/synchronization/ConnectionBroadcastReceiver.java
hb0: This was old code which I moved. I now dropped the null check. If this can actually occur, I think the SDK user forgot to remove the listener.

PR #65, opened 2018-09-13
Task/cy 4080 space warnining strategies

hb0:


datacapturing/src/main/java/de/cyface/datacapturing/EventHandlingStrategy.java
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Interface for strategies to respond to events triggered by the {@link DataCapturingBackgroundService}.
+ * E.g.: Show a notification when little space is available and stop the capturing.
+ * Must be {@link Parcelable} to be passed from the {@link DataCapturingService} via {@link Intent}.
+ *
+ * @author Armin Schnabel
+ * @version 1.0.0
+ * @since 2.5.0
+ */
```
hb0: I named this Strategy Interface quite general to avoid frequent API changes on the DataCapturingService constructor. This way we hopefully only have to pass this Strategy class no matter how many events we add to this interface to handle.

datacapturing/src/main/java/de/cyface/datacapturing/backend/DataCapturingBackgroundService.java
```diff
@@ -214,6 +223,12 @@ public int onStartCommand(final Intent intent, final int flags, final int startI
             final String authority = intent.getCharSequenceExtra(AUTHORITY_ID).toString();
             persistenceLayer = new MeasurementPersistence(this.getContentResolver(), authority);
 
+            // Loads EventHandlingStrategy
+            this.eventHandlingStrategy = intent.getParcelableExtra(EVENT_HANDLING_STRATEGY_ID);
+            if (eventHandlingStrategy == null) {
+                this.eventHandlingStrategy = new IgnoreEventsStrategy();
```
hb0: Not sure if it's cleaner to allow library users to pass null if they don't care about those events or if we want to force them to pass the IgnoreEventsStrategy themselves if they don't want to implement a strategy themselves.

muthenberg: It would be the cleanest if there are two constructors. One requiring the EventHandlingStrategy and another one using the default.

datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingListener.java
```diff
@@ -67,4 +67,9 @@
      * @return <code>true</code> if the permission was granted; <code>false</code> otherwise.
      */
     boolean onRequiresPermission(String permission, Reason reason);
+
+    /**
+     * This method is called when the capturing stopped.
```
muthenberg: Maybe provide a little bit of explanation here on when this might occur. (Such as space is low).

datacapturing/src/cyface/java/de/cyface/datacapturing/CyfaceDataCapturingService.java
```diff
@@ -31,12 +33,15 @@
...
      *            this service.
+     * @param eventHandlingStrategy The {@link EventHandlingStrategy} used to react to selected events
+     *                              triggered by the {@link DataCapturingBackgroundService}.
      * @throws SetupException If writing the components preferences or registering the dummy user account fails.
      */
     public CyfaceDataCapturingService(final @NonNull Context context, final @NonNull ContentResolver contentResolver,
             final @NonNull String authority, final @NonNull String accountType,
-            final @NonNull String dataUploadServerAddress) throws SetupException {
-        super(context, contentResolver, authority, accountType, dataUploadServerAddress);
+            final @NonNull String dataUploadServerAddress, final @Nullable EventHandlingStrategy eventHandlingStrategy)
```
muthenberg: See my comment in the discussion.

PR #67, opened 2018-09-17
Subtask/cy 4082 fix tests capturing services

hb0:
You already accepted the parent branch changes but I cannot merge it back to dev because:
- tests had to be fixed (this PR)
- the dev (submodules) has to be rebased - I'll do this after this PR is completed

As the DataCapturingServiceTests were not completely indepent (at least one test forgot to shutdown the DataCapturingService and there was no check that this cannot happen) and as some used sync and some the async interface I had to refactore the test class to see what the tests are actually doing.

Thus, this PR is a bit longer.

### DIFF
--- a/datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
+++ b/datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
@@ -34,7 +34,6 @@ import android.os.Messenger;
 import android.os.RemoteException;
 import android.preference.PreferenceManager;
 import android.support.annotation.NonNull;
-import android.support.annotation.Nullable;
 import android.support.v4.app.ActivityCompat;
 import android.util.Log;
 
@@ -184,7 +183,7 @@ public abstract class DataCapturingService {
      */
     public DataCapturingService(final @NonNull Context context, final @NonNull ContentResolver resolver,
             final @NonNull String authority, final @NonNull String accountType,
-            final @NonNull String dataUploadServerAddress, final @Nullable EventHandlingStrategy eventHandlingStrategy)
+            final @NonNull String dataUploadServerAddress, final @NonNull EventHandlingStrategy eventHandlingStrategy)
             throws SetupException {
         this.context = new WeakReference<>(context);
         this.authority = authority;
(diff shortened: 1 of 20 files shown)


datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
```diff
@@ -791,28 +790,72 @@ private void stopServiceSync(final long timeout, final @NonNull TimeUnit unit)
...
 
+    /**
+     * This message is sent to the {@link ShutDownFinishedHandler} to inform callers that the async stop
+     * command was executed and that the {@link DataCapturingBackgroundService} was not active when the
+     * stop command was received which can happen when the pause command was executed before.
+     *
+     * @param context The {@link Context} used to send the broadcast from
+     */
+    private void sendServiceStoppedUnsuccessfullyCallback(final Context context) {
+        sendServiceStoppedBroadcast(context, -1, false);
```
muthenberg: If we stop a paused service, there should still have been an open measurement (which should be closed by the current call to stop) and thus we should no have to use -1 here.

datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
```diff
@@ -791,28 +790,72 @@ private void stopServiceSync(final long timeout, final @NonNull TimeUnit unit)
...
+     */
+    private void sendServiceStoppedBroadcast(final Context context, final long measurementIdentifier,
+                                             final boolean stoppedSuccessfully) {
+        final Intent stoppedBroadcastIntent = new Intent(MessageCodes.FINISHED_HANDLER_BROADCAST_SERVICE_STOPPED);
+        if (stoppedSuccessfully) {
+            Validate.isTrue(measurementIdentifier > 0L);
+            stoppedBroadcastIntent.putExtra(MEASUREMENT_ID, measurementIdentifier);
+        }
+        stoppedBroadcastIntent.putExtra(STOPPED_SUCCESSFULLY, stoppedSuccessfully);
+        context.sendBroadcast(stoppedBroadcastIntent);
```
muthenberg: You need to send a local broadcast here or else all apps using the SDK will receive the broadcast (Stadtradeln, Cyface, Cyface Pro).

PR #71, opened 2018-09-20
Bug/cy 4090 space warning notification not shown

hb0:


datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
```diff
@@ -1164,11 +1164,10 @@ public void handleMessage(final @NonNull Message msg) {
                         final Condition condition = lock.newCondition();
                         final StopSynchronizer synchronizationReceiver = new StopSynchronizer(lock, condition);
                         try {
-                            // To make sure the background service is stopped, we unbind this service
-                            // from it via the stopService method (to reduce code duplicity). As the
-                            // background service stopped itself in advance, we expect no active service:
-                            Validate.isTrue(!dataCapturingService.stopService(new Measurement(measurementId),
-                                    synchronizationReceiver));
```
hb0: I had to remove the Validation as the background service was still running (still bound) so the app would crash.

datacapturing/src/main/java/de/cyface/datacapturing/ui/CapturingNotification.java
```diff
@@ -18,11 +18,17 @@
...
+ * @version 2.1.1
  * @since 1.0.0
  */
 public class CapturingNotification {
 
+    /**
+     * Identifies the capturing ongoing {@link android.app.Notification} which can be implemented by sdk using apps.
+     * These ids should be placed centrally to avoid duplicate entries.
+     */
+    public final static int CAPTURING_NOTIFICATION_ID = 1;
```
muthenberg: Wollten wir es nicht dem User überlassen, das hier zu übergeben?

datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
muthenberg: Shouldn't the validation work in that case? (stopService returns true since the service was still running and was successfully stopped?)

datacapturing/src/main/java/de/cyface/datacapturing/ui/CapturingNotification.java
muthenberg: Never mind. This is the legacy way for the old notification id, I see! Maybe add a todo, so I do not get confused.

datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
hb0: Well, it should return true because it's already in the stop-universe and unbound by the stopService command. But I didn't dare to validate for this ;-) I'll add the validation again and check if it works locally and add a test to ensure we don't break it later on. (As we currently don't test the spaceWarning strategy implemented by our app.) To be continued....

hb0: manuelly tested: works

PR #75, opened 2018-10-09
Replace capturing stopped global broadcast with local broadcast #CY-4181

hb0:
Only the CAPTURING_STOPPED global broadcast was replaced by a local broadcast as the other global broadcasts are inter-process communication and, thus, don't work with local broadcasts (tested).

But I added the task number and deprecated tag to those broadcasts and some docu.

Most of the code changes are renamings and docu, the only functional difference is that the STOPPED broadcast is now locally un-/register/sent

### DIFF
--- a/datacapturing/src/androidTest/java/de/cyface/datacapturing/backend/BackgroundServiceTest.java
+++ b/datacapturing/src/androidTest/java/de/cyface/datacapturing/backend/BackgroundServiceTest.java
@@ -170,7 +170,7 @@ public class BackgroundServiceTest {
             @Override
             public void run() {
                 PongReceiver isRunningChecker = new PongReceiver(context);
-                isRunningChecker.pongAndReceive(2, TimeUnit.SECONDS, testCallback);
+                isRunningChecker.asyncIsRunningCheck(2, TimeUnit.SECONDS, testCallback);
             }
         });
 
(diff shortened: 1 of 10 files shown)


datacapturing/src/androidTest/java/de/cyface/datacapturing/backend/BackgroundServiceTest.java
```diff
@@ -170,7 +170,7 @@ public void testStartDataCapturingTwice() throws InterruptedException, TimeoutEx
             @Override
             public void run() {
                 PongReceiver isRunningChecker = new PongReceiver(context);
-                isRunningChecker.pongAndReceive(2, TimeUnit.SECONDS, testCallback);
```
hb0: Solves a TODO from you as this method was named wrongly

muthenberg: Ok. but method names should always be verb phrases. In this case maybe 'checkIsRunningAsync' or something along the lines.

datacapturing/src/main/java/de/cyface/datacapturing/MessageCodes.java
```diff
@@ -1,11 +1,16 @@
...
 
+import de.cyface.datacapturing.backend.DataCapturingBackgroundService;
+
 /**
- * This class is a wrapper for all message codes used by the Cyface backend to send inter process communication (IPC)
- * messages.
+ * This class is a wrapper for all message codes used by the Cyface backend to send inner- and inter process
+ * communication (IPC) messages. For safety and efficiency we use LocalBroadcasts for inner process communication
+ * as they are very easy to use and {@link android.os.Message}s for inter-process communication as the
+ * LocalBroadcasts can't be passed between processes.
```
muthenberg: Do you think this documentation is appropriately placed here? This gives away quite some implementation details. Javadoc should be more about how to use a class and not about how it is implemented. These facts are of no interest to me, if I need to work with it. 
I guess we should put that under implementation advice in the readme (which probably does not exist yet).

PR #77, opened 2018-11-02
Bug/cy 3575 sdk integrating apps not independent

hb0:
all inter process messages are sent via global broadcast with prefix and service stopped message (inner process) as local broadcast


datacapturing/src/main/java/de/cyface/datacapturing/PongReceiver.java
```diff
@@ -25,7 +25,7 @@
  * tell the caller, that the service is not running.
  *
  * @author Klemens Muthmann
- * @version 1.1.4
+ * @version 1.1.7
```
muthenberg: You should add your name here.
```suggestion
* @author Armin Schnabel
 * @version 1.1.7
```

PR #80, opened 2019-01-09
Epos/mov 371 file storage fixed

hb0:
Schlappe ~8000 Zeilen Diff. Einige Änderungen sollten dir bereits bekannt vorkommen, genau sagen kann ich das leider nicht da ich nicht weiß, wie genau du dich schon mit dem neuen Code auseinander  gesetzt hast im notification und measurement view branch.
Sobald das hier in den dev gemergt wurde würde ich die Alpha Version für Phillip und uns zum testen releasen und wieder am FILES Meilenstein arbeiten.


persistence/src/main/java/de/cyface/persistence/Persistence.java
```diff
@@ -0,0 +1,682 @@
...
+/**
+ * This class wraps the Cyface Android persistence API as required by the <code>DataCapturingService</code>,
+ * <code>SyncAdapter</code> and its
+ * delegate objects.
+ *
+ * @author Armin Schnabel
+ * @version 1.0.2
+ * @since 3.0.0
+ */
+public class Persistence {
```
muthenberg: I really dislike the idea that you can create instances of this and also instances of the subclass `MeasurementPersistence`. Creating instances of non-leaf classes (with respect to the inheritance tree) is really confusing.

persistence/src/main/java/de/cyface/persistence/Persistence.java
```diff
@@ -0,0 +1,682 @@
...
+     */
+    private final ContentResolver resolver;
+    /**
+     * The authority used to identify the Android content provider to persist data to or load it from.
+     */
+    private final String authority;
+    /**
+     * The {@link File} pointing to the directory containing open measurements.
+     */
+    protected final File openMeasurementsDir;
```
muthenberg: Don't you have wrapper classes like MeasurementFile, which you should use here?

persistence/src/main/java/de/cyface/persistence/Persistence.java
```diff
@@ -0,0 +1,682 @@
...
+                : new byte[] {};
+
+        final ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+        outputStream.write(transferFileFormat);
+        outputStream.write(pointCounts);
+        outputStream.write(geoLocationData);
+        outputStream.write(accelerationData);
+        outputStream.write(rotationData);
+        outputStream.write(directionData);
+        return outputStream.toByteArray();
```
muthenberg: Hopefully this does not crash for large measurements.

persistence/src/main/java/de/cyface/persistence/serialization/AccelerationsFile.java
```diff
@@ -0,0 +1,81 @@
...
+import de.cyface.persistence.model.Point3D;
+
+/**
+ * The file format to persist the captured acceleration points.
+ *
+ * @author Armin Schnabel
+ * @version 1.0.0
+ * @since 3.0.0
+ */
+public class AccelerationsFile implements FileSupport<List<Point3D>> {
```
muthenberg: Why are there classes for all the different sensors. Shouldn't one class be enough? They behave equally except for the location they store the data to, so this difference should be solvable using parameters.

persistence/src/main/java/de/cyface/persistence/serialization/MetaFile.java
```diff
@@ -0,0 +1,216 @@
...
+     * The name of the file containing the data
+     */
+    public static final String FILE_EXTENSION = "cyfm";
+
+    /**
+     * @param context The {@link Context} required to access the persistence layer.
+     * @param measurementId The identifier of the measurement
+     * @param vehicle The {@link Vehicle} used in the measurement
+     */
+    public MetaFile(final Context context, final long measurementId, final Vehicle vehicle) {
```
muthenberg: The `context` should be NonNull

persistence/src/main/java/de/cyface/persistence/FileUtils.java
```diff
@@ -0,0 +1,215 @@
...
+    /**
+     * Returns the path of the parent directory containing all folders of corrupted measurements.
+     * This directory is deleted when the app is uninstalled and can only be accessed by the app.
+     */
+    private String corruptedMeasurementsDirPath;
+
+    /**
+     * @param context The {@link Context} required to locate the app's internal storage directory.
+     */
+    public FileUtils(final Context context) {
```
muthenberg: If you remove that context and add the files dir as a parameter you could test the whole class using simple unit tests, avoiding any Android shenanigans.

persistence/src/main/java/de/cyface/persistence/serialization/MetaFile.java
```diff
@@ -0,0 +1,216 @@
...
+        final byte[] data = serialize(metaData);
+        append(data, this.file);
+    }
+
+    private static void append(final byte[] data, final File file) {
+        try {
+            final BufferedOutputStream outputStream = new BufferedOutputStream(new FileOutputStream(file, true));
+            Log.d(TAG, "Writing " + data.length + " bytes (4 counters)");
+            outputStream.write(data);
+            outputStream.close();
```
muthenberg: This results in a memory leak if the output stream fails while writing the data. Use finally!

persistence/src/main/java/de/cyface/persistence/serialization/MetaFile.java
```diff
@@ -0,0 +1,216 @@
...
+    /**
+     * To resume a paused measurement this method loads the point counters from the {@link MetaFile}
+     * and removes the counters from it so the updates counts can be stored when finishing or pausing
+     * the measurement later on (again).
+     *
+     * @param context The {@link Context} required to access the persistence layer.
+     * @param measurementId The identifier of the measurement to resume
+     * @return the {@link MetaData} containing the point counters
+     */
+    public static MetaData resume(final Context context, final long measurementId) {
```
muthenberg: context is non null?

persistence/src/main/java/de/cyface/persistence/serialization/MetaFile.java
```diff
@@ -0,0 +1,216 @@
...
+        dataFormatVersionBytes[1] = (byte)PERSISTENCE_FILE_FORMAT_VERSION;
+        final byte[] vehicleIdBytes = MeasurementSerializer.serialize(vehicle);
+
+        try {
+            final BufferedOutputStream outputStream = new BufferedOutputStream(new FileOutputStream(file, false));
+            Log.d(TAG, "Writing " + dataFormatVersionBytes.length + " bytes (File Format)");
+            outputStream.write(dataFormatVersionBytes);
+            Log.d(TAG, "Writing " + vehicleIdBytes.length + " bytes (vehicle id)");
+            outputStream.write(vehicleIdBytes);
+            outputStream.close();
```
muthenberg: memory leak

persistence/src/main/java/de/cyface/persistence/serialization/RotationsFile.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    public void append(final List<Point3D> dataPoints) { // was: Serializable not DataPoint
+
+        final byte[] data = serialize(dataPoints);
+
+        try {
+            final BufferedOutputStream outputStream = new BufferedOutputStream(new FileOutputStream(file, true));
+            outputStream.write(data);
+            outputStream.close();
```
muthenberg: memory leak

synchronization/src/androidTestCyface/java/de/cyface/synchronization/CyfaceAuthenticatorTest.java
```diff
@@ -14,33 +13,29 @@
...
  * @author Armin Schnabel
- * @version 1.1.0
+ * @version 1.1.2
  * @since 2.0.0
  */
 
 @RunWith(AndroidJUnit4.class)
 @LargeTest
-@FlakyTest
+@FlakyTest // Flaky means (because of build.gradle) that this test is not executed in the Mock flavour (because it
```
muthenberg: The way it is done now we could probably write our own annotation and base the exclusion on that annotation. That way we could use a more intuitive name.

synchronization/src/androidTestCyface/java/de/cyface/synchronization/UploadProgressTest.java
```diff
@@ -17,55 +19,47 @@
...
 import android.accounts.AccountManager;
-import android.content.BroadcastReceiver;
-import android.content.ContentProviderClient;
-import android.content.ContentResolver;
-import android.content.Context;
-import android.content.Intent;
-import android.content.IntentFilter;
-import android.content.SharedPreferences;
-import android.content.SyncResult;
+import android.content.*;
```
muthenberg: no wildcard imports

synchronization/src/androidTestCyface/java/de/cyface/synchronization/UploadProgressTest.java
```diff
@@ -1,7 +1,9 @@
 package de.cyface.synchronization;
 
 import static de.cyface.synchronization.CyfaceConnectionStatusListener.SYNC_PERCENTAGE;
+import static de.cyface.synchronization.SharedConstants.DEVICE_IDENTIFIER_KEY;
 import static de.cyface.synchronization.TestUtils.*;
```
muthenberg: no wildcard imports

synchronization/src/androidTestMovebis/java/de/cyface/synchronization/DataTransmissionTest.java
```diff
@@ -112,14 +120,13 @@ public void testUploadSomeBytesViaMultiPart()
                 throw new IllegalStateException(
                         String.format("Unable to acquire client for content provider %s", AUTHORITY));
 
-            MeasurementContentProviderClient loader = new MeasurementContentProviderClient(measurementIdentifier,
-                    client, AUTHORITY);
-            MeasurementSerializer serializer = new MeasurementSerializer();
-            InputStream measurementData = serializer.serialize(loader);
+            InputStream measurementData = persistence.loadSerializedCompressed(measurementIdentifier);
```
muthenberg: You should close this stream somewhere in a finally block.

synchronization/src/cyface/java/de/cyface/synchronization/CyfaceAuthenticator.java
```diff
@@ -125,7 +134,7 @@ public Bundle getAuthToken(final @Nullable AccountAuthenticatorResponse response
 
                 // Get Auth token
                 try {
-                    authToken = initSync(account.name, password, sslContext);
+                    authToken = initSync(account.name, password, sslContext); // HOW TO GET THE AUTHORITY HERE ?
```
muthenberg: How was it done previously?

synchronization/src/main/java/de/cyface/synchronization/SharedConstants.java
```diff
@@ -2,10 +2,11 @@
 
 /**
  * Final static constants used by multiple classes.
- * FIXME: until tests run through again I wont merge the constants of both flavours, thus we have  a 3rd shared const file
+ * FIXME: until tests run through again I wont merge the constants of both flavours, thus we have a 3rd shared const
+ * file
  *
  * @author Armin Schnabel
- * @version 1.0.0
+ * @version 1.0.3
```
muthenberg: I would say this is version 2.0.0

testutils/build.gradle
```diff
@@ -0,0 +1,55 @@
...
+    // - If Junit symbols are not resolvable in IntelliJ, make sure Build Variant is set to debug
+    // - Loading another dependency (e.g. module) only it's production dependencies (compile) are loaded but not other dependencies (e.g. testCompile)
+    //implementation 'com.android.support:appcompat-v7:' + rootProject.ext.junitVersion - auto generated by Android Studio
+    implementation "junit:junit:$rootProject.ext.junitVersion"
+    implementation "org.mockito:mockito-core:$rootProject.ext.mockitoVersion"
+    // Optional - For better debuggable asserts
+    //implementation "org.hamcrest:hamcrest-all:$rootProject.ext.hamcrestVersion"
+
+    // Dependencies for instrumentation tests
+    // Force usage of support annotations in the test app, since it is internally used by the runner module.
```
muthenberg: This comment seems wrong now.

testutils/build.gradle
```diff
@@ -0,0 +1,55 @@
...
+
+}
+
+dependencies {
+    implementation "androidx.annotation:annotation:$rootProject.ext.androidxAnnotationVersion"
+
+    // Dependencies for local unit tests
+    // - If Junit symbols are not resolvable in IntelliJ, make sure Build Variant is set to debug
+    // - Loading another dependency (e.g. module) only it's production dependencies (compile) are loaded but not other dependencies (e.g. testCompile)
+    //implementation 'com.android.support:appcompat-v7:' + rootProject.ext.junitVersion - auto generated by Android Studio
```
muthenberg: This dependency should be removed. It is not valid under AndroidX.

testutils/build.gradle
```diff
@@ -0,0 +1,55 @@
...
+    // Optional - For better debuggable asserts
+    //implementation "org.hamcrest:hamcrest-all:$rootProject.ext.hamcrestVersion"
+
+    // Dependencies for instrumentation tests
+    // Force usage of support annotations in the test app, since it is internally used by the runner module.
+    implementation "androidx.test.ext:junit:$rootProject.ext.extJunitVersion"
+    implementation "androidx.test:rules:$rootProject.ext.rulesVersion"
+
+    implementation project(":utils")
+    implementation project(path: ':persistence')
```
muthenberg: Either use `path` or not.

datacapturing/src/androidTest/java/de/cyface/datacapturing/backend/BackgroundServiceTest.java
```diff
@@ -1,6 +1,6 @@
 package de.cyface.datacapturing.backend;
 
-import static de.cyface.datacapturing.BundlesExtrasCodes.EVENT_HANDLING_STRATEGY_ID;
+import static de.cyface.datacapturing.BundlesExtrasCodes.*;
```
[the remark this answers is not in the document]
hb0: My IDE decided since the last update that in the auto-format this is the way to go -.- Don't ask me why, I assumed you changes something in the eclipse auto-formatter file but on the other hand: I did not update the repository in which this file is located.

datacapturing/src/androidTest/java/de/cyface/datacapturing/persistence/PersistenceTest.java
```diff
@@ -0,0 +1,252 @@
...
+        byte[] decompressedBytes = new byte[1000];
+        int decompressedLength = inflater.inflate(decompressedBytes);
+        inflater.end();
+        assertThat(decompressedLength, is(equalTo(SERIALIZED_SIZE)));
+        assertThat(Arrays.copyOfRange(decompressedBytes, 0, SERIALIZED_SIZE), is(equalTo(serializedData)));
+    }
+
+    private long insertSerializationTestSample() throws NoSuchMeasurementException {
+        // Insert sample measurement data
+        Persistence persistence = new Persistence(context, resolver, AUTHORITY);
```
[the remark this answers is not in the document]
hb0: @muthenberg 
**Persistence**
- this is used to access the persistence layer, e.g. to create a new measurement in tests or to load data from the persistence layer or to load the compressed and serialized data for synchronization, to delete a measurement, mark it as synced, etc. Thus, it's used to access the persistence layer after the capturing took place (not live capturing).

**MeasurementPersistence**
- extends Persistence for the live-capturing (used in DataCapturingService + DCBackgroundService). E.g.: It reuses the Persistence.new Meaurement() method but makes sure that the currentMeasurementIdentifier is cached. It also offers the DataCapturingService to closeRecentMeasurement(), store the newly captured location or sensor data and it starts and shut downs a threadpool for the capturing.

**So maybe we could rename MeasurementPersistence to DataCapturing(Service)Persistence ? Any better suggestions?**

hb0:
Implemented second part of the comments in #dfa16219c2db8097810e8567d41ba0971cc3a7a7

datacapturing/src/androidTestCyface/java/de/cyface/datacapturing/DataCapturingServiceTest.java
```diff
@@ -51,11 +45,12 @@
  *
  * @author Klemens Muthmann
  * @author Armin Schnabel
- * @version 4.2.2
+ * @version 4.2.4
  * @since 2.0.0
  */
 @RunWith(AndroidJUnit4.class)
-@MediumTest
+@FlakyTest
```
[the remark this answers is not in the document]
hb0: Not sure where this came from ... I remove all flaky flags from this class. Thanks for the catch.

datacapturing/src/androidTestCyface/java/de/cyface/datacapturing/DataCapturingServiceTest.java
```diff
@@ -516,8 +529,8 @@ public void testStartPauseStop()
     /**
      * Tests if the service lifecycle is running successfully.
      * <p>
-     * Makes sure the {@link DataCapturingService#pauseAsync(ShutDownFinishedHandler)} and
-     * {@link DataCapturingService#resumeAsync(StartUpFinishedHandler)}
+     * Makes sure the {@link DataCapturingService#pauseAsync(ShutDownFinishedHandler)} ()} and
+     * {@link DataCapturingService#resumeAsync(StartUpFinishedHandler)} ()}
```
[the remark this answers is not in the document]
hb0: ide bug or user keyboard input bug. Resolved.

datacapturing/src/androidTestCyface/java/de/cyface/datacapturing/DataCapturingServiceTestWithoutPermission.java
```diff
@@ -46,25 +47,36 @@
...
+     * Condition waiting for the background service to wake up this test case.
+     */
+    private Condition condition;
+
     /**
      * Initializes the object of class under test.
      */
     @Before
     public void setUp() {
         final Context context = InstrumentationRegistry.getTargetContext();
```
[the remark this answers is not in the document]
hb0: This test is set to @Ignore - see code: 
`@Ignore // Ignore this test until Android is capable of resetting permissions for every test`

I tried to execute it but in testUIListenerIsInformedOfMissingPermission()
```
assertThat(exceptionCaught, is(equalTo(true)));
```
is false and testServiceDoesNotStartWithoutPermission() does not throw the MissingPermissionException.

If you think this test should work again and must be fixed an enabled please create yourself a task to do so as I did not add the @Ignore flag in this PR so I think it should not block this PR.

datacapturing/src/main/java/de/cyface/datacapturing/BundlesExtrasCodes.java
```diff
@@ -33,6 +39,30 @@
...
+     * Code that contains the number of entries stored in the {@link RotationsFile} captured. This allows the
+     * {@link DataCapturingService} to keep the number of points stored when a measurement was paused and to resume the
+     * counting by passing it back to the {@link DataCapturingBackgroundService} on resume.
+     */
+    public static final String ROTATION_POINT_COUNT = "de.cyface.extra.rotation_point.count";
+    /**
+     * Code that contains the number of entries stored in the {@link DirectionsFile} captured. This allows the
+     * {@link DataCapturingService} to keep the number of points stored when a measurement was paused and to resume the
+     * counting by passing it back to the {@link DataCapturingBackgroundService} on resume.
+     */
```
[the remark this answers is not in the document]
hb0: Beim Pausieren der Messung werden die counter in die MetaFile persistiert.
Beim Resumen werden sie aus der File geladen (und entfernt damit klar ist, dass die Messung wieder aktiv und nicht pausiert ist). Es folgt sofort der Start des BackgroundServices bei dem über das start Bundle die Counter übertragen werden. D.h. in dem Bundle sind die Counter nur ein paar Millisekunden gespeichert. Persistent sind die in der MetaFile gespeichert. Befindet sich die Messung mit gespeicherten Counter im open folder bedeutet das pausierte Messung, ohne counter bedeutet das, dass das eine laufende messung ist (z.b. Resumed). Und mit Counter und in finished/ liegend wurde die Messung korrekt beendet.
Dadurch können wir gestartete aber gestorbene Messungen beim nächsten Start erkennen (und von pausierten unterscheiden) und als corrupted markieren und ggf. reparieren oder löschen.

datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
```diff
@@ -214,42 +195,13 @@ public DataCapturingService(final @NonNull Context context, final @NonNull Conte
...
     /**
      * Starts the capturing process with a listener, that is notified of important events occurring while the capturing
      * process is running.
      * <p>
      * This method returns as soon as starting the service was initiated. You may not assume the service is running,
      * after the method returns. Please use the {@link StartUpFinishedHandler} to receive a callback, when the service
-     * has been started or use the synchronized version {@link #startSync(DataCapturingListener, Vehicle)}.
+     * has been started.
      * <p>
      * This method is thread safe to call.
```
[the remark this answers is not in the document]
hb0: Please create a task for this. Out of the blue I'm not sure what exactly you mean but this does not seem to have anything to do with the changes suggested in this PR.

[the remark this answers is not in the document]
hb0: Maybe you mean this task? https://cyface.atlassian.net/browse/CY-4098

datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
```diff
@@ -467,36 +360,72 @@ public void resumeSync() throws DataCapturingException, MissingPermissionExcepti
...
     }
 
+    /**
+     * Returns open measurements currently on this device.
+     *
+     * @return A list containing all open measurements currently stored on this device by this application. An empty
+     *         list if
+     *         there are no such measurements, but never <code>null</code>.
+     */
+    public @NonNull List<Measurement> loadOpenMeasurements() {
```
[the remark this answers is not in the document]
hb0: We could do this I suppose (this helps to ensure these methods work correctly). We still need to observe the measurements folder ofc but could load the new state on each update.

Please add a task for this - as you were the last one who worked the view.

datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
```diff
@@ -526,18 +456,18 @@ public void forceMeasurementSynchronisation(final @NonNull String username) thro
...
-    // TODO provide a custom list implementation that loads only small portions into memory.
-
     /**
      * Loads a track of geo locations for an existing {@link Measurement}. This method loads the complete track into
      * memory. For large tracks this could slow down the device or even reach the applications memory limit.
      *
      * @param measurement The <code>measurement</code> to load all the geo locations for.
      * @return The track associated with the measurement as a list of ordered (by timestamp) geo locations.
+     *         /
+     *         public List<GeoLocation> loadTrack(final @NonNull Measurement measurement) throws
```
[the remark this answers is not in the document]
hb0: Ok I comment this in again and document that this method must be supported for the movebis project. There is a simple test for the underlying method (persistence.loadTrack) (CapturedDataWriterTest#testLoadTrack). If you think we need another, specific test please create a task with the expected test content.

datacapturing/src/main/java/de/cyface/datacapturing/persistence/MeasurementPersistence.java
```diff
@@ -2,118 +2,100 @@
...
     /**
      * Creates a new completely initialized <code>MeasurementPersistence</code>.
      *
      * @param resolver <code>ContentResolver</code> that provides access to the {@link MeasuringPointsContentProvider}.
      * @param authority The authority used to identify the Android content provider to persist data to or load it from.
+     * @param context The {@link Context} required to locate the app's internal storage directory.
      */
-    public MeasurementPersistence(final @NonNull ContentResolver resolver, final @NonNull String authority) {
-        this.resolver = resolver;
+    public MeasurementPersistence(@NonNull final Context context, final @NonNull ContentResolver resolver,
```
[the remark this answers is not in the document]
hb0: Ok, my suggestion: I think it's intuitive to:
* first use the @ flags (because the also are usually the first thing before a method declaration?)
* then final/static (just because ... alphabetical order - I don't know how else to decide this?)

datacapturing/src/main/java/de/cyface/datacapturing/persistence/MeasurementPersistence.java
```diff
@@ -127,9 +109,18 @@ public void storeData(final @NonNull CapturedData data, final long measurementId
...
+        if (rotationsFile == null) {
+            rotationsFile = new RotationsFile(context, measurementIdentifier);
+        }
+        if (directionsFile == null) {
+            directionsFile = new DirectionsFile(context, measurementIdentifier);
+        }
 
-        CapturedDataWriter writer = new CapturedDataWriter(data, resolver, authority, measurementIdentifier, callback);
-
+        CapturedDataWriter writer = new CapturedDataWriter(data, accelerationsFile, rotationsFile, directionsFile,
```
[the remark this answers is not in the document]
hb0: Good question. Can we test it? Or should we create a spike task to read about this first ? Who should be assigned? Would you mind to create the task because I think you have more knowledge/experience here ...?

datacapturing/src/main/java/de/cyface/datacapturing/persistence/MeasurementPersistence.java
```diff
@@ -140,291 +131,47 @@ public void storeData(final @NonNull CapturedData data, final long measurementId
...
-
-    /**
-     * Loads the track of <code>GeoLocation</code> objects for the provided measurement.
-     *
-     * @param measurement The measurement to load the track for.
-     * @return The loaded track of <code>GeoLocation</code> objects ordered by time ascending.
-     *
-     * @throws NoSuchMeasurementException If the provided measurement was <code>null</code>.
-     */
-    public List<GeoLocation> loadTrack(final @NonNull Measurement measurement) throws NoSuchMeasurementException {
```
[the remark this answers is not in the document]
hb0: I commented the uncommented interface method of the DataCapturingService in again. This should be resolved by that I think?

datacapturing/src/main/java/de/cyface/datacapturing/persistence/MeasurementPersistence.java
```diff
@@ -459,39 +204,36 @@ public void shutdown() {
...
-        return new Uri.Builder().scheme("content").authority(authority).appendPath(DirectionPointTable.URI_PATH)
-                .build();
+     * Before starting or resuming a measurement we need to make sure that there are no corrupted measurements
+     * left in the /open/ folder. This can happen when the {@link DataCapturingBackgroundService} dies a devastating
+     * Exception-death and was not able to append the {@link MetaFile.PointMetaData} to the {@link MetaFile}.
+     *
+     * FIXME: If we want to support double start/resume calls this method needs to be adjusted
+     * so that it does not clean open measurements when there is capturing going on, else we kill
+     * an ongoing measurement! We could also make sure this method is not called when capturing is
+     * taking place. But how do we synchronize this? The capturing could be in progress of starting,
```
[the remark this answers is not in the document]
hb0: I removed this fixme in a following PR as this test works again (I added some simple code which ignores a second call). But it's probably not thread safe. Thus, I'll add this comment to the life-cycle methods on handler task: https://cyface.atlassian.net/browse/CY-4098

hb0:
Implemented 3rd part of the comments in #7461abc68826f9e0ea2299eeab7e27cc9b123dae

persistence/src/main/java/de/cyface/persistence/Persistence.java
hb0: If I understand you correctly:

The Constructor of the **non-leaf Persistence class** is used in the SyncAdapter.onPerformSync() in order to load the measurements from the persistence layer, to serialize them and to mark them as synced.

1) If we would use the **leaf class MeasurementPersistence** then a threadpool (used for capturing) would be created which we don't need, which is the only real difference between the constructors.

2) The other difference is that MeasurementPersistence caches a reference to the currentlyCapturedMeasurement which should not really be a problem for the synchronization as the synchronization does not create a new Measurement (which is where the cache is set) so it would just be null all the time which is ok I guess?

**Any suggestions how we could handle this differently in a clean way?**

The name of this class is discussed in a comment above.

persistence/src/main/java/de/cyface/persistence/Persistence.java
hb0: it's now in the Measurement class and FileUtils (to generate the names)

persistence/src/main/java/de/cyface/persistence/Persistence.java
hb0: I added a TODO to test this on weak devices - all FIXMEs must be removed before release and all TODOs are checked and annotated with a Task before final release.

persistence/src/main/java/de/cyface/persistence/serialization/AccelerationsFile.java
hb0: True. taking care of it right now...

persistence/src/main/java/de/cyface/persistence/FileUtils.java
hb0: Alread done. Had to google "shenanigans" though :jack_o_lantern:

synchronization/src/androidTestCyface/java/de/cyface/synchronization/CyfaceAuthenticatorTest.java
hb0: Interesting idea. Put it into a task if you fancy so ;-)

synchronization/src/androidTestCyface/java/de/cyface/synchronization/UploadProgressTest.java
hb0: I searched and replaced all wildcard imports in the code

synchronization/src/cyface/java/de/cyface/synchronization/CyfaceAuthenticator.java
hb0: deprecated comment of mine, ignore it (removed it)

persistence/src/main/java/de/cyface/persistence/Persistence.java
muthenberg: The thing is that clean code requires classes to either be designed for inheritance. In that case it should not be possible to create instances and required to create a child with a concrete implementation. So if there really is a difference we should do something like:

Persistence
|-> NormalPersistence
|-> MeasurementPersistence

If the difference is very superficial we should think about our architecture. Do we really need this differentiation? Is there a place where you provide a reference to a Persistence which can be filled by either NormalPersistence or MeasurementPersistence? If there is not, we should prefer delegation over inheritance. So in that case we move everything that is different between MeasurementPersistence and NormalPersistence to two new classes inheriting from a common ancestor and provide a reference to that ancestor to the constructor of Persistence (MeasurementPersistence and NormalPersistence would not exist in that case). I hope this is kind of clear. In the end we would have:

PersistenceLayer (no subclasses)
- <<attribute>> behaviour: PersistenceBehaviour
+ <<constructor>> PersistenceLayer(PersistenceBehaviour)

PersistenceBehaviour (please think of a proper name)
|-> NormalPersistenceBehaviour
| -> MeasurementPersistenceBehaviour

But you do have better insight if such a setup is even necessary.

hb0: The behavioral pattern is such a beauty here ...

PR #82, opened 2019-01-24
Stad 10 placeholder notification

muthenberg:


datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
```diff
@@ -1,3 +1,21 @@
+/*
```
hb0: (Almost) nothing changes in this file but all got re-ordered so git things the code was deleted and a few lines later that (the old code) was newly added. Please make sure the order of the methods is not changes or else it's gonna be a pain to merge this with the 10k lines or code PR which comes next.

muthenberg: Should be fixed now.

PR #88, opened 2019-02-06
Task/mov 530 reenable used apis for sr and document usages

hb0:
This is a stacked PR (ontop of the other) because:
* this task is urgent (red)
* I branched from the open PR branch and did not want to cherry-pick if I don't have to

I.e. after reviewing the other PR you can review the changes here so I can merge this back a bit quicker.

The changes here are mostly documentation of API methods and enabling of uncommented api methods used in the SR code.

### DIFF
--- a/datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
+++ b/datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
@@ -51,6 +51,7 @@ import android.os.Messenger;
 import android.os.RemoteException;
 import android.preference.PreferenceManager;
 import android.util.Log;
+
 import androidx.annotation.NonNull;
 import androidx.core.app.ActivityCompat;
 import androidx.localbroadcastmanager.content.LocalBroadcastManager;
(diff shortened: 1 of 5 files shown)


datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
```diff
@@ -587,8 +581,8 @@ public void disconnect() throws DataCapturingException {
      *
      * @throws IllegalStateException If communication with background service is not successful.
      */
-    @SuppressWarnings("WeakerAccess") // because we need to support this API - TODO really?
-    public void reconnect() throws DataCapturingException {
+    @SuppressWarnings("WeakerAccess") // TODO: not used by RS. Do we use it in our app?
```
muthenberg: This should be used by all apps to rebind the app after it has been shut down.

hb0: added this as TODO there so that I'll use this in our app and add sample code to the Readme

hb0:
I'll create a new PR for this branch into dev for a clean git history

PR #91, opened 2019-02-07
Task/mov 521 typed persistence layer

hb0:
Wie gewünscht


hb0:
Die Dokumentation muss ich vermutlich noch anpassen.

datacapturing/src/androidTest/java/de/cyface/datacapturing/persistence/PersistenceLayerTest.java
```diff
@@ -49,7 +49,7 @@
     /**
      * An object of the class under test. It is setup prior to each test execution.
      */
-    private PersistenceLayer oocut;
+    private PersistenceLayer<CapturingPersistenceBehaviour> oocut;
```
hb0: CapturingPersistenceBehaviour because in this test workflows are tested which are used during capturing, e.g. lead finished measurements while one is still running, etc.

PR #103, opened 2019-02-20
Task/mov 565 release current fixes for rm

hb0:
The changes in this PR contain the latest fixes. This should make life for RM easier when working with an unstable SDK version.


hb0:
from target branch

PR #113, opened 2019-03-14
Bug/mov 609 sync without wifi

hb0:
Leichtes Refactoring das zum einen:
- [einen Bug aus dem Plan.io behebt](https://stadtradeln.plan.io/issues/1475?issue_count=10&issue_position=3&next_issue_id=1470&prev_issue_id=1474)
- und zum anderen Methoden und Variablen semantisch korrekter benennt
- und alle mir bekannten sync bugs oder tests die mir eingefallen sind erfolgreich absolviert

Mal sehen wie sich das auf anderen Handys/Nutzer bzw. Mittelzeit Tests schlägt.

Das "instantly cancel ongoing sync when syncable network is gone" wie heute im daily erwähnt habe ich nun ins Backlog da das v.a. Resourcen schont für einen Spezialfall (serialization für größere Messung startet, Wifi wird disconnected, aktuell läuft die Serialization dann weiter aber führt keinen Upload durch weil davor in onPerformSync geprüft wird, ob denn noch eine syncable connection besteht.

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,10 +321,7 @@ public class MainFragment extends Fragment implements ConnectionStatusListener {
             return;
         }
         
-        // Login and create account 
-        // a) Static token variant:
-        dataCapturingService.registerJWTAuthToken(username, token);
-        // or b) Login via LoginActivity and using dynamic tokens
+        // Login via LoginActivity, create account and using dynamic tokens
         // The LoginActivity is called by Android which handles the account creation
         accountManager.addAccount(ACCOUNT_TYPE, AUTH_TOKEN_TYPE, null, null,
             getMainActivityFromContext(context), new AccountManagerCallback<Bundle>() {
--- a/datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
+++ b/datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
@@ -485,7 +485,7 @@ public abstract class DataCapturingService {
      */
     @SuppressWarnings({"WeakerAccess", "unused"}) // Used by implementing app (CY)
     public void scheduleSyncNow() {
-        getWiFiSurveyor().scheduleSyncNow();
+        surveyor.scheduleSyncNow();
     }
 
     /**
@@ -867,11 +867,11 @@ public abstract class DataCapturingService {
     }
 
     /**
-     * see {@link WiFiSurveyor#isConnected()}
+     * see {@link WiFiSurveyor#isConnectedToSyncableNetwork()}
      */
     @SuppressWarnings("unused") // Allows implementing apps (CY) to only trigger sync manually when connected
     public boolean isConnected() {
-        return getWiFiSurveyor().isConnected();
+        return surveyor.isConnectedToSyncableNetwork();
     }
 
     /**
--- a/datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
+++ b/datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
@@ -92,7 +92,7 @@ import de.cyface.utils.Validate;
  *
  * @author Klemens Muthmann
  * @author Armin Schnabel
- * @version 14.0.3
+ * @version 14.0.4
  * @since 1.0.0
  */
 public abstract class DataCapturingService {
--- a/datacapturing/src/movebis/java/de/cyface/datacapturing/MovebisDataCapturingService.java
+++ b/datacapturing/src/movebis/java/de/cyface/datacapturing/MovebisDataCapturingService.java
@@ -53,7 +53,7 @@ import de.cyface.utils.CursorIsNullException;
  *
  * @author Klemens Muthmann
  * @author Armin Schnabel
- * @version 8.0.0
+ * @version 8.0.1
  * @since 2.0.0
  */
 @SuppressWarnings({"unused", "WeakerAccess"}) // Sdk implementing apps (SR) use to create a DataCapturingService
--- a/datacapturing/src/movebis/java/de/cyface/datacapturing/MovebisDataCapturingService.java
+++ b/datacapturing/src/movebis/java/de/cyface/datacapturing/MovebisDataCapturingService.java
@@ -225,6 +225,9 @@ public class MovebisDataCapturingService extends DataCapturingService {
             throws SynchronisationException {
         final AccountManager accountManager = AccountManager.get(getContext());
 
+        // The workflow here is slightly different from the one in CyfaceDataCapturingService.
+        // If problems are reported, ensure they are exactly the same (i.e. reuse existing account).
+
         // Create a "dummy" account used for auto synchronization. Null password as the token is static
         final Account synchronizationAccount = getWiFiSurveyor().createAccount(username, null);
 
--- a/synchronization/src/androidTestCyface/java/de/cyface/synchronization/SyncAdapterTest.java
+++ b/synchronization/src/androidTestCyface/java/de/cyface/synchronization/SyncAdapterTest.java
@@ -2,6 +2,7 @@ package de.cyface.synchronization;
 
 import static de.cyface.persistence.Utils.getGeoLocationsUri;
 import static de.cyface.persistence.Utils.getMeasurementUri;
+import static de.cyface.synchronization.SyncAdapter.MOCKED_IS_PERIODIC_SYNC_DISABLED_FALSE;
 import static de.cyface.synchronization.TestUtils.ACCOUNT_TYPE;
 import static de.cyface.synchronization.TestUtils.AUTHORITY;
 import static de.cyface.synchronization.TestUtils.TEST_API_URL;
@@ -16,9 +17,10 @@ import java.util.List;
 
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Ignore;
+import org.junit.FixMethodOrder;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.junit.runners.MethodSorters;
 
 import android.accounts.Account;
 import android.accounts.AccountManager;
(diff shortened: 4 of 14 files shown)


README.md
```diff
@@ -321,10 +321,7 @@ public class MainFragment extends Fragment implements ConnectionStatusListener {
             return;
         }
         
-        // Login and create account 
-        // a) Static token variant:
-        dataCapturingService.registerJWTAuthToken(username, token);
```
hb0: because this is in the "CyfaceDCS only" section so there should not be any mentions of the MovebisDCS code usage

datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
```diff
@@ -867,11 +867,11 @@ void shutdownConnectionStatusReceiver() {
     }
 
     /**
-     * see {@link WiFiSurveyor#isConnected()}
+     * see {@link WiFiSurveyor#isConnectedToSyncableNetwork()}
      */
     @SuppressWarnings("unused") // Allows implementing apps (CY) to only trigger sync manually when connected
     public boolean isConnected() {
-        return getWiFiSurveyor().isConnected();
+        return surveyor.isConnectedToSyncableNetwork();
```
hb0: Is semantically clearer as there is a "isConnected" method in Network which returns true when the network is connected, but we want this method to only return true when there is a syncable network connected (i.e. wifi only? then only when such a network is connected)

datacapturing/src/movebis/java/de/cyface/datacapturing/MovebisDataCapturingService.java
```diff
@@ -225,6 +225,9 @@ public void registerJWTAuthToken(final @NonNull String username, final @NonNull
             throws SynchronisationException {
         final AccountManager accountManager = AccountManager.get(getContext());
```
hb0: I added this in case there are problems which we cannot reproduce in our app. We probably do not need to change anything so I didn't want to start refactoring when there is no need.

synchronization/src/androidTestMovebis/java/de/cyface/synchronization/SyncAdapterTest.java
```diff
@@ -58,7 +58,7 @@ public void testRequestSync() throws InterruptedException {
         Account newAccount = new Account(TestUtils.DEFAULT_USERNAME, ACCOUNT_TYPE);
         if (am.addAccountExplicitly(newAccount, TestUtils.DEFAULT_PASSWORD, Bundle.EMPTY)) {
             ContentResolver.setIsSyncable(newAccount, AUTHORITY, 1);
-            ContentResolver.setSyncAutomatically(newAccount, AUTHORITY, true);
+            ContentResolver.addPeriodicSync(newAccount, AUTHORITY, Bundle.EMPTY, 1);
```
hb0: The combination of syncAutomatically + periodic sync did not work as we expected which lead to the sync starting when though there is no wifi network. We now only use syncPeriodically and when it's activated (e.g. just connected to wifi) we trigger a initial sync which is what the user expects.

synchronization/src/cyface/java/de/cyface/synchronization/CyfaceAuthenticator.java
```diff
@@ -277,11 +277,12 @@ public Bundle hasFeatures(AccountAuthenticatorResponse response, Account account
      * @throws ResponseParsingException When the http response could not be parsed.
      * @throws SynchronisationException When the new data output for the http connection failed to be created.
      * @throws UnauthorizedException If the credentials for the cyface server are wrong.
+     * @throws NetworkErrorException when the connection's input or error stream was null (e.g. connection disabled)
```
hb0: this error lead to a crash when I enabled wifi and disabled it very fast as soon as it was connected. This captures it softly as all other sync errors

synchronization/src/main/java/de/cyface/synchronization/Http.java
```diff
@@ -63,10 +65,11 @@ HttpsURLConnection openHttpConnection(@NonNull URL url, @NonNull SSLContext sslC
      * @throws ResponseParsingException When the http response could not be parsed.
      * @throws SynchronisationException When the new data output for the http connection failed to be created.
      * @throws UnauthorizedException If the credentials for the cyface server are wrong.
+     * @throws NetworkErrorException when the connection's input or error stream was null
      */
     HttpResponse post(HttpURLConnection connection, JSONObject payload, boolean compress)
             throws RequestParsingException, DataTransmissionException, SynchronisationException,
-            ResponseParsingException, UnauthorizedException, BadRequestException;
+            ResponseParsingException, UnauthorizedException, BadRequestException, NetworkErrorException;
```
hb0: Is this also seen as an API change? I guess so ... thus the code version bump

synchronization/src/main/java/de/cyface/synchronization/WiFiSurveyor.java
```diff
@@ -60,11 +64,6 @@
      * The current Android context (i.e. Activity or Service).
      */
     private WeakReference<Context> context;
-    /**
-     * A flag that might be queried to see whether synchronization is active or not. This is <code>true</code> if
-     * synchronization is active and <code>false</code> otherwise.
-     */
-    private boolean synchronizationIsActive;
```
hb0: This attribute was misleading. As long as we do not directly cancel synchronization when we change the auto-sync settings this would say there is no active sync while there may be still an ongoing active sync (e.g. serialization).
There is a new task to add cancelation but as I see it it's not so urgent right now.

hb0: This attribute was misleading. As long as we do not directly cancel synchronization when we change the auto-sync settings this would say there is no active sync while there may be still an ongoing active sync (e.g. serialization).
There is a new task to add cancelation but as I see it it's not so urgent right now.

synchronization/src/androidTestCyface/java/de/cyface/synchronization/SyncAdapterTest.java
```diff
@@ -51,24 +52,43 @@
...
  * @author Armin Schnabel
  * @author Klemens Muthmann
- * @version 2.2.3
+ * @version 2.2.4
  * @since 2.4.0
  */
 @RunWith(AndroidJUnit4.class)
+// To execute notice errors with the short running testOnPerformSync before the large
+// testOnPerformSyncWithLargeMeasurement which can save a lot of time
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
hb0: when we execute the test from the IDE the large test (~7-10 minutes) would be executed before the small one. Usually I only want to execute the small one as I know that the large one is executed on the CI anyway and it only tests that no OOM is produced during upload which should seldom break.

So this is a fast fix to execute the small test first and then I can cancel the other one if I want to.

synchronization/src/androidTestCyface/java/de/cyface/synchronization/SyncAdapterTest.java
```diff
@@ -108,7 +123,10 @@ public void testOnPerformSync() throws NoSuchMeasurementException, CursorIsNullE
             client = contentResolver.acquireContentProviderClient(getGeoLocationsUri(AUTHORITY));
             final SyncResult result = new SyncResult();
             Validate.notNull(client);
-            syncAdapter.onPerformSync(account, new Bundle(), AUTHORITY, client, result);
+
+            final Bundle testBundle = new Bundle();
+            testBundle.putString(MOCKED_IS_PERIODIC_SYNC_DISABLED_FALSE, "");
```
hb0: see onPerformSync in SyncAdapter. Seems like a more or less clean mock injection as I cannot change the onPerformSync interface.

synchronization/src/main/java/de/cyface/synchronization/SyncAdapter.java
```diff
@@ -42,13 +43,23 @@
...
  * @since 2.0.0
  */
 public final class SyncAdapter extends AbstractThreadedSyncAdapter {
 
+    /**
+     * This bundle flag allows our unit tests to mock isPeriodicSyncDisabled(). We cannot use addPeriodicSync() as we do
+     * in the production code as this is an Unit test. When this {@code Bundle} extra is set (no matter to which String)
+     * the {@link #isPeriodicSyncDisabled(Account, String)} method returns false;
+     */
+    static final String MOCKED_IS_PERIODIC_SYNC_DISABLED_FALSE = "mocked_periodic_sync_check_false";
```
hb0: as I explained above

synchronization/src/androidTestCyface/java/de/cyface/synchronization/SyncAdapterTest.java
muthenberg: Yeah. The usual Android problem. I know. Everything involving the Android core framework is not very testable.

synchronization/src/cyface/java/de/cyface/synchronization/CyfaceAuthenticator.java
muthenberg: That is probably the error Manuel told me about today.

synchronization/src/main/java/de/cyface/synchronization/Http.java
muthenberg: Correct. Checked Exceptions are always an API change.

synchronization/src/main/java/de/cyface/synchronization/HttpConnection.java
```diff
@@ -265,9 +274,11 @@ private String addPart(final @NonNull String key, final @NonNull String value, f
      * @throws ResponseParsingException If the system fails in handling the HTTP response.
      * @throws UnauthorizedException If the credentials for the cyface server are wrong.
      * @throws SynchronisationException If a connection error occurred while reading the response code.
+     * @throws NetworkErrorException when the connection's input or error stream was null
      */
-    private HttpResponse readResponse(final @NonNull HttpURLConnection con) throws ResponseParsingException,
-            DataTransmissionException, UnauthorizedException, SynchronisationException, BadRequestException {
+    private HttpResponse readResponse(final @NonNull HttpURLConnection con)
```
muthenberg: Didn't we say some time we would move the Annotation in front of the final?

hb0: We did. I didnt add this line now but just removed or added exceptions so in this special case I did not clean up all old lines such as this. But I'm doing this very often these days - changing the order correctly.

PR #126, opened 2019-03-25
Remove validation which should not have been in production code

hb0:
The ValidationException thrown could be the cause that the synchronization does not start on the STAD app so I release this as fix until I can reproduce the bug locally.

### DIFF
--- a/synchronization/src/main/java/de/cyface/synchronization/WiFiSurveyor.java
+++ b/synchronization/src/main/java/de/cyface/synchronization/WiFiSurveyor.java
@@ -44,7 +44,7 @@ import de.cyface.utils.Validate;
  *
  * @author Klemens Muthmann
  * @author Armin Schnabel
- * @version 7.0.0
+ * @version 7.0.1
  * @since 2.0.0
  */
 public class WiFiSurveyor extends BroadcastReceiver {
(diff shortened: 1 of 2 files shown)


synchronization/src/main/java/de/cyface/synchronization/WiFiSurveyor.java
```diff
@@ -336,22 +336,7 @@ public void makeAccountSyncable(@NonNull final Account account, boolean enabled)
...
-     *
-     * @param account The {@code Account} to be checked.
-     */
-    public static void validateAccountFlags(@NonNull final Account account, @NonNull final String authority) {
-        final boolean periodicSyncRegistered = ContentResolver.getPeriodicSyncs(account, authority).size() > 0;
-        final boolean autoSyncEnabled = ContentResolver.getSyncAutomatically(account, authority);
-        Validate.isTrue(periodicSyncRegistered == autoSyncEnabled,
-                "Both, periodicSync and autoSync, must be in the same state but are: " + periodicSyncRegistered
-                        + " and " + autoSyncEnabled + ", in this order");
+            // Do not use validateAccountFlags in production code as periodicSync flags are set async
```
hb0: This is the important change in the production code:
- As the periodicSync flag is changed asynchronously we must not check that the change was successful immediately. This throws an exception in the STAD app. For the same reason I moved the validation method to the test flavour to make sure we don't accidentally  use this again in production.
- there is a WifiSurveyorTest which validates that the periodicSync flag is set

hb0: As the public validateAccountFlags method was not used by others and is not used anymore I did not change the 1st digit code version for this file.

PR #131, opened 2019-03-27
Bug/mov 527 data capturing service tests flaky

hb0:
**Bug was:**
- DataCapturingServiceTests are flaky

**Symptoms were**
- there was a race condition in those tests (for details see MOV-527)
- because of this the ping receiver was sometimes registered after the ping was sent

**Cause was**
- the PingReceiver was registered in BGS.onStartCommand
- the Test is using this call order: startService, bindService, PongReceiver.checkIsRunning()
- bindService() is released as soon as onStartCommand is called but does not wait for it to be finished
- and there we have our race condition - the PingReceiver is sometimes registered slower in onStartCommand than the Test's PongReceiver.checkIsRunning() was sending the Ping itself so the PingReceiver missed the Ping because of it's slow start

**Fix is**
- onBind() is called before onStartCommand so I wanted to move the PingReceiver registration there
- we used the deviceId to make the global broadcasts used app-unique (so that multiple SDK implementing apps don't influence each other)
- now we can't use the deviceId as global broadcast suffix as we have to load it (since recently) from the persistence Layer which requires the authority which we cannot pass to onBind() in the BGS via intent as we do in onStartCommand because the Android documentation (and the reality) says intent extras are ignored!
- for this reason we use the app's package name which only requires the context and is perfect to differentiate between apps on the same device.

**Other Notes**
- I know that it would be cleaner to avoid global broadcasts but I tried this before (high time effort) but failed using local broadcasts or Messengers for ping/pong.
- I have a subtask MOV-659 to make sure SDK implementing apps still don't influence each other with the new app-unique global broadcast suffix. I also checked via debugging that package name != sdk package but of the app (in the integration test it was de.cyface.datacapturing.test)


persistence/src/androidTest/AndroidManifest.xml
```diff
@@ -1,13 +1,13 @@
...
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
-    package="de.cyface.persistence">
+    package="de.cyface.persistence.test">
 
     <!-- The integration test uses a different authority to be executed next to the default app (same device) -->
     <application>
         <provider
             android:name="de.cyface.persistence.MeasuringPointsContentProvider"
-            android:authorities="de.cyface.provider.test"
```
hb0: see MOV-636

datacapturing/src/main/java/de/cyface/datacapturing/backend/DataCapturingBackgroundService.java
```diff
@@ -150,11 +151,33 @@
...
+     * <p>
+     * <b>Attention:</b> The identifier must be identical in the global broadcast sender and receiver.
+     */
+    private String appId;
 
     @Override
     public IBinder onBind(final @NonNull Intent intent) {
         Log.v(TAG, String.format("Binding to %s", this.getClass().getName()));
+
+        // FIXME [MOV-660]: Is onBind() called more often than onStartCommand()? If so, we need to make sure Ping is
```
hb0: Hier bin ich mir gerade (langer Tag, spät Abends) nicht mehr sicher:

**Wird vom BGS erwartet, dass er einen PingReceiver jedes mal neu stellt bei jedem bind (siehe unten) und diesen registriert?
Können wir hierdurch mehrere PingReceiver haben was schlecht wäre?**

bindService() is called in:
1. DCS.bind() and this is called by:
1.1 DCS.reconnect()
1.2 DCS.runService() - pingReceiver expected
2. DCS.onBindingDied() to rebind

datacapturing/src/main/java/de/cyface/datacapturing/backend/DataCapturingBackgroundService.java
```diff
@@ -188,14 +193,30 @@ public void onCreate() {
...
+            Log.v(TAG, "onBind: Ping Receiver was already registered");
+            return;
+        }
+
+        // Allows other parties to ping this service to see if it is running
+        // We cannot use the deviceId as device-unique app identifier as we need the authority (persistence) for this
+        // which we cannot pass via bind() as documented by the {@link #onBind()} method.
+        final String appId = getBaseContext().getPackageName();
+        pingReceiver = new PingReceiver(appId);
+        registerReceiver(pingReceiver, new IntentFilter(MessageCodes.getPingActionId(appId)));
```
hb0: Moveing this to here is basically the core fix of the bug.

synchronization/src/main/java/de/cyface/synchronization/HttpConnection.java
```diff
@@ -161,15 +161,17 @@ public HttpResponse post(@NonNull final HttpURLConnection connection, final @Non
...
-        final long requestLength = header.getBytes().length + filePartSize;
-        // This should be obsolete with setFixedLengthStreamingMode:
-        // connection.setRequestProperty("Content-length", String.valueOf(requestLength));
-        connection.setFixedLengthStreamingMode((int)requestLength);
+        // Attention: Parts of the header (Content-Type, boundary, request method, user agent) are already set
+        final String remainingHeader = generateHeader(metaData, fileName);
+
+        // Set the fixed number of bytes which will be written to the OutputStream
+        final long binarySize = transferTempFile.length();
+        final long bytesWrittenToOutputStream = calculateBytesWrittenToOutputStream(remainingHeader, binarySize);
```
hb0: I have this in a method to test this in a unit test to make sure we don't break it again.

PR #132, opened 2019-04-02
Bug/mov 669 wrong binary header

hb0:
**Bug was:**
- the sensor point counters in the transfer binary header were larger than the number of sensor points in the binary body

**Possible Cause**
- we did not ensure that the counters were only increased when the data was successfully written to the binary (atomic)

**Fix**
- To avoid complex code we now don't count the data during capturing but count the points stored in the binaries just before we sync the data based on the actual number of binaries which were written


datacapturing/src/androidTest/java/de/cyface/datacapturing/model/CapturedDataWriterTest.java
```diff
@@ -173,9 +173,6 @@ public void testCreateNewMeasurement() throws NoSuchMeasurementException, Cursor
                     is(equalTo(Vehicle.UNKNOWN.getDatabaseIdentifier())));
             assertThat(result.getString(result.getColumnIndex(MeasurementTable.COLUMN_STATUS)),
                     is(equalTo(MeasurementStatus.OPEN.getDatabaseIdentifier())));
-            assertThat(result.getInt(result.getColumnIndex(MeasurementTable.COLUMN_ACCELERATIONS)), is(equalTo(0)));
```
hb0: we don't store the sensor point counters in the measurement entry (database) anymore but just calculate them before synchronization in the temporary transfer binary

datacapturing/src/main/java/de/cyface/datacapturing/backend/DataCapturingBackgroundService.java
```diff
@@ -139,10 +137,6 @@
      * The strategy used to calculate the {@link Measurement#distance} from {@link GeoLocation} pairs
      */
     DistanceCalculationStrategy distanceCalculationStrategy;
-    /**
-     * Meta information required for deserialization of {@link Point3dFile}s.
-     */
-    PointMetaData pointMetaData;
```
hb0: those are the points counters we don't count/store during capturing anymore

persistence/src/main/java/de/cyface/persistence/serialization/MeasurementSerializer.java
```diff
@@ -306,8 +311,6 @@ private void loadSerializedCompressed(@NonNull final OutputStream fileOutputStre
      * <b>ATTENTION:</b> The caller must make sure the {@param bufferedOutputStream} is closed when no longer needed
      * or the app crashes.
      *
-     * TODO [MOV-554]: test performance on weak devices for large measurements
```
hb0: this was done in the linked task (finished)

persistence/src/main/java/de/cyface/persistence/serialization/MeasurementSerializer.java
```diff
@@ -356,6 +355,31 @@ public void loadSerialized(@NonNull final BufferedOutputStream bufferedOutputStr
...
                 Point3dFile.DIRECTIONS_FOLDER_NAME, Point3dFile.DIRECTION_FILE_EXTENSION);
 
+        // Calculate how many points the files contain (for the binary header)
+        int accelerationsCount = 0;
+        int rotationsCount = 0;
+        int directionsCount = 0;
+        // noinspection ConstantConditions // can happen in tests
+        if (accelerationFile != null && accelerationFile.exists()) {
+            accelerationsCount = (int)(accelerationFile.length() / BYTES_IN_ONE_POINT_3D_ENTRY);
+            Validate.isTrue(accelerationsCount * BYTES_IN_ONE_POINT_3D_ENTRY == accelerationFile.length());
```
hb0: To make sure there is no "half" point entry at the end (i.e. bytes % bytes_per_point == 0)

This way we see if this happens at all so that we need to handle this.

persistence/src/test/java/de/cyface/persistence/DatabaseHelperTest.java
```diff
@@ -112,7 +112,41 @@ public void tearDown() {
...
+     */
+    @Test
+    public void testMigrationV12ToV13() {
+
+        // Arrange
+        // This is simpler than copying and adjusting the code from previous versions
+        createV11DatabaseWithData(db);
+        oocut.onUpgrade(db, 11, 12);
+
+        // Act - This is how the method is called by the system (not incrementally!)
```
hb0: That's just a reminder for future tests that the system calls this method not incrementally, i.e. if we want to rest V10->V13 migration, we should call (db, 10, 13) to ensure we test the code realistically.

In this case we actually just want to test V12-V13, the V11-V12 migration before is part of the "Arrange". We avoid copy/pasting a sample dataset for V12 this way by reusing the V11 sample data set.

synchronization/src/androidTest/java/de/cyface/synchronization/WifiSurveyorTest.java
```diff
@@ -97,6 +98,7 @@ public void tearDown() {
      * This test may be flaky on a read device when the network changes during the test.
      */
     @Test
+    @Ignore // TODO [MOV-644]: flaky
```
hb0: I've this task "in doing" but postponed because of the open priority bugs. This way I don't have to execute the CI build 5+ times until this test goes through -.-

synchronization/src/test/java/de/cyface/synchronization/MeasurementSerializerTest.java
```diff
@@ -100,20 +102,26 @@
...
+    @Mock
+    private File mockedRotationFile;
+    @Mock
+    private File mockedDirectionFile;
     private MeasurementSerializer oocut;
-
-    private final static long SERIALIZED_SIZE = BYTES_IN_HEADER + 3L * BYTES_IN_ONE_GEO_LOCATION_ENTRY
-            + 3L * 3L * BYTES_IN_ONE_POINT_3D_ENTRY;
-
-    private final static int SAMPLE_POINT_3DS = 3;
```
hb0: I did not like that we generate the same amount of AP/RP/DP as we might not catch all bugs, e.g. referencing RP instead of AP somewhere in the code. With different sensor point type counts in the sample sets we should notice such typos with those tests.

synchronization/src/test/java/de/cyface/synchronization/MeasurementSerializerTest.java
```diff
@@ -217,10 +283,12 @@ public void testDeserializeMeasurement() throws CursorIsNullException, IOExcepti
      *
      */
     @Test
-    public void testSerializeCompressedMeasurement() throws CursorIsNullException, IOException {
+    public void testSerializeCompressedMeasurement() throws CursorIsNullException {
 
-        final long SERIALIZED_COMPRESSED_SIZE = 24L; // When compression Deflater(level 9, true)
-        final File compressedTransferBytes = oocut.writeSerializedCompressed(loader, 0, persistence);
+        // If you need to change the sample point counts (and this) make sure the test work with the previous counts
+        final long SERIALIZED_COMPRESSED_SIZE = 43L; // When compression Deflater(level 9, true)
```
hb0: Before changing this number I checked that the new code still ends up this 24 bytes compressed data when using 3/3/3 sensor points. Then I changed the numbers to 21/20/10 and hard-coded the new compressed byte count into this test here.

synchronization/src/test/java/de/cyface/synchronization/MeasurementSerializerTest.java
```diff
@@ -234,14 +302,14 @@ public void testDecompressDeserialize() throws IOException, DataFormatException,
...
-        final byte[] compressedBytes = new byte[(int) compressedTransferTempFile.length()];
+        final byte[] compressedBytes = new byte[(int)compressedTransferTempFile.length()];
         DataInputStream dis = new DataInputStream(new FileInputStream(compressedTransferTempFile));
         dis.readFully(compressedBytes);
 
         // Decompress the compressed bytes and check length and bytes
         Inflater inflater = new Inflater(COMPRESSION_NOWRAP);
         inflater.setInput(compressedBytes, 0, compressedBytes.length);
-        byte[] decompressedBytes = new byte[1000];
+        byte[] decompressedBytes = new byte[2000];
```
hb0: because now the serialized bytes are ~ 1750 which would be cut when we still use a 1000 Byte array.

PR #142, opened 2019-04-12
Bug/mov 650 no sync on wifi android 6

hb0:
I tried to make the commits atomar reviewable, let me know which aspects are not. Thanks.

### DIFF
--- a/synchronization/src/main/java/de/cyface/synchronization/NetworkCallback.java
+++ b/synchronization/src/main/java/de/cyface/synchronization/NetworkCallback.java
@@ -30,7 +30,7 @@ import de.cyface.utils.Validate;
 
 /**
  * This callback handles status changes of the {@link Network} connectivity, e.g. to determine if synchronization should
- * be enabled depending on the {@link android.net.NetworkCapabilities#NET_CAPABILITY_NOT_METERED} capabilities of the
+ * be enabled depending on the {@code NetworkCapabilities#NET_CAPABILITY_NOT_METERED} capabilities of the
  * newly connected network.
  *
  * @author Armin Schnabel
--- a/synchronization/src/main/java/de/cyface/synchronization/NetworkCallback.java
+++ b/synchronization/src/main/java/de/cyface/synchronization/NetworkCallback.java
@@ -14,6 +14,7 @@
  */
 package de.cyface.synchronization;
 
+import static de.cyface.synchronization.WiFiSurveyor.MINIMUM_VERSION_TO_USE_NOT_METERED_FLAG;
 import static de.cyface.synchronization.WiFiSurveyor.TAG;
 
 import android.accounts.Account;
@@ -33,7 +34,7 @@ import de.cyface.utils.Validate;
  * newly connected network.
  *
  * @author Armin Schnabel
- * @version 2.0.1
+ * @version 2.1.0
  * @since 3.0.0
  */
 @TargetApi(Build.VERSION_CODES.LOLLIPOP)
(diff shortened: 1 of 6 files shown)


synchronization/src/main/java/de/cyface/synchronization/NetworkCallback.java
```diff
@@ -58,12 +59,8 @@ public void onCapabilitiesChanged(@NonNull final Network network, @NonNull final
...
         // Ensure this event is only triggered for not metered connections when syncOnUnMeteredNetworkOnly
         if (surveyor.isSyncOnUnMeteredNetworkOnly()) {
-            final boolean notMetered = capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_NOT_METERED);
-            // This should always be the case but for some reasons it's not (Nexus 5)
-            if (!notMetered) {
-                Log.e(TAG, "onCapabilitiesChanged called on metered network with isSyncOnUnMeteredNetworkOnly on");
-                return;
-            }
+            final boolean unMeteredNetwork = capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_NOT_METERED);
+            Validate.isTrue(unMeteredNetwork);
```
muthenberg: Doesn't this crash everything if we are on a metered network?

synchronization/src/main/java/de/cyface/synchronization/WiFiSurveyor.java
```diff
@@ -336,7 +335,7 @@ public void makeAccountSyncable(@NonNull final Account account, boolean enabled)
         ContentResolver.setSyncAutomatically(account, authority, false);
         setSyncEnabled(account, enabled);
 
-            // Do not use validateAccountFlags in production code as periodicSync flags are set async
+        // Do not use validateAccountFlags in production code as periodicSync flags are set async
```
muthenberg: Comment has no context. What is validateAccountFlags?

synchronization/src/main/java/de/cyface/synchronization/NetworkCallback.java
```diff
@@ -54,13 +55,28 @@
...
         // Ensure this event is only triggered for not metered connections when syncOnUnMeteredNetworkOnly
         if (surveyor.isSyncOnUnMeteredNetworkOnly()) {
-            final boolean unMeteredNetwork = capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_NOT_METERED);
-            Validate.isTrue(unMeteredNetwork);
+
+            final boolean isUsingUnMeteredCheckInsteadOfWifi = Build.VERSION.SDK_INT >= MINIMUM_VERSION_TO_USE_NOT_METERED_FLAG;
+            if (isUsingUnMeteredCheckInsteadOfWifi) {
+                Validate.isTrue(capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_NOT_METERED));
+            } else {
+                Validate.isTrue(capabilities.hasTransport(NetworkCapabilities.TRANSPORT_WIFI));
```
muthenberg: This still crashes everything on a connection loss, doesn't it?

hb0: No, because we only register to NetworkCallbacks with those attributes (see below).

Those validations just makes sure this "Filter" actually works as expected because I had a bug before where I accidentally used another filter in the callback registration than what I was expecting in here:

Here (or in the code below) the Registration:
```
final NetworkRequest.Builder requestBuilder = new NetworkRequest.Builder();
            if (syncOnUnMeteredNetworkOnly) {	          
                Log.v(TAG, "startSurveillance for wifi networks only");	                
               final boolean isUsingUnMeteredCheckInsteadOfWifi = Build.VERSION.SDK_INT >= MINIMUM_VERSION_TO_USE_NOT_METERED_FLAG;	
                  if (isUsingUnMeteredCheckInsteadOfWifi) {
                    requestBuilder.addCapability(NET_CAPABILITY_NOT_METERED);
                } else {
                    requestBuilder.addTransportType(NetworkCapabilities.TRANSPORT_WIFI);
                }
```

synchronization/src/main/java/de/cyface/synchronization/WiFiSurveyor.java
hb0: That's a method I moved into test code. Had this validation in here before when I still thought that periodicSyncFlags are set synchroniously. I'll pushed a commit which clarifies this comment.

PR #143, opened 2019-04-16
Add valiation to ensure the same number of bytes are written to the s…

hb0:
…tream as registered

### DIFF
--- a/synchronization/src/main/java/de/cyface/synchronization/HttpConnection.java
+++ b/synchronization/src/main/java/de/cyface/synchronization/HttpConnection.java
@@ -43,13 +43,14 @@ import android.os.Build;
 import android.util.Log;
 
 import androidx.annotation.NonNull;
+import de.cyface.utils.Validate;
 
 /**
  * Implements the {@link Http} connection interface for the Cyface apps.
  *
  * @author Klemens Muthmann
  * @author Armin Schnabel
- * @version 6.0.0
+ * @version 6.0.1
  * @since 2.0.0
  */
 public class HttpConnection implements Http {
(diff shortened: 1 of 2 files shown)


synchronization/src/main/java/de/cyface/synchronization/HttpConnection.java
```diff
@@ -194,9 +197,11 @@ public HttpResponse post(@NonNull final HttpURLConnection connection, final @Non
         try {
             connection.connect();
             try {
+                long bytesWrittenToOutputStream = 0;
+
                 // Write MultiPart header (including the filePartHeader
-                outputStream.write(remainingHeader.getBytes());
-                outputStream.flush();
```
hb0: explicit flush is not necessary

synchronization/src/main/java/de/cyface/synchronization/HttpConnection.java
```diff
@@ -208,24 +213,30 @@ public HttpResponse post(@NonNull final HttpURLConnection connection, final @Non
...
                 }
+                bytesWrittenToOutputStream += bytesWritten;
 
                 // Write MultiPart Tail boundary
                 outputStream.write(TAIL.getBytes());
+                bytesWrittenToOutputStream += TAIL.getBytes().length;
+
+                // Ensure we only write the "registered" number of bytes to the stream MOV-693
+                outputStream.flush(); // This way we can identify exceptions thrown by flush easier
+                Validate.isTrue(bytesWrittenToOutputStream == fixedStreamLength, "bytesWrittenToOutputStream "
```
hb0: This way I want to ensure the problem is not that the wrong number of bytes to be written to the stream is registered.

hb0: It's currently also crashing, so no change here.

PR #152, opened 2019-04-29
Fix Packaging for build Variants

cirk2:
At some point the output name of the build .aar Files changed. This refactors the publishing config to publish all flavours using their configuration.


hb0:
Declined, da ich das im PR #153 behandle.

PR #159, opened 2019-05-10
Bug/mov 723 multiple exceptions on logout

hb0:


synchronization/src/main/java/de/cyface/synchronization/SyncAdapter.java
```diff
@@ -124,6 +125,12 @@ public void onPerformSync(@NonNull final Account account, @NonNull final Bundle
         // This allows us to mock the #isConnected() check for unit tests
         mockIsConnectedToReturnTrue = extras.containsKey(MOCK_IS_CONNECTED_TO_RETURN_TRUE);
 
+        // If WifiSurveyor.shutdownSurveillance was called in the meantime, cancel sync:
```
hb0: We only want to have the hard ValidationException below if the sync was not interrupted by us before (e.g. logout), which is a the only legitimate case where the token may be null.

synchronization/src/main/java/de/cyface/synchronization/WiFiSurveyor.java
```diff
@@ -233,6 +233,10 @@ public void stopSurveillance() throws SynchronisationException {
                 throw new SynchronisationException(e);
             }
         }
+
+        // This interrupts ongoing synchronization or else it may crash because required data is missing (token,
+        // account)
+        ContentResolver.cancelSync(currentSynchronizationAccount, authority);
```
hb0: sets the interrupted() flag to true

synchronization/src/main/java/de/cyface/synchronization/WiFiSurveyor.java
```diff
@@ -226,6 +226,7 @@ public void stopSurveillance() throws SynchronisationException {
                 return;
             }
             connectivityManager.unregisterNetworkCallback(networkCallback);
+            networkCallback = null;
```
hb0: forgot this before. This prevents a crash when you log out (which unregisters the network callbacks) and kill the app, which also unregister the callback if it's not null.

datacapturing/src/cyface/java/de/cyface/datacapturing/CyfaceDataCapturingService.java
```diff
@@ -154,6 +155,25 @@ public void startWifiSurveyor() throws SetupException {
         }
     }
 
+    /**
```
hb0: that's the same code as MovebisDataCapturingService.deregisterJwtToken executes, which was what I wanted to test in our app to reproduce and fix bugs. That's why I added the logout functionality to our app, which calls this method here.

synchronization/src/test/java/de/cyface/synchronization/WiFiSurveyorTest.java
```diff
@@ -271,7 +271,14 @@ private void setConnectivity(final boolean connected, final boolean wifiNotMobil
         // Now we can call the NetworkCallbacks with the correct networkCapabilities (registration is ok)
         // Only call the networkCallback for wifi connections as we only register those in production code
         if (wifiNotMobile || !surveyor.isSyncOnUnMeteredNetworkOnly()) {
-            Validate.isTrue(shadowConnectivityManager.getNetworkCallbacks().size() == 1);
+            final boolean isAccountRegistered = oocut.currentSynchronizationAccount != null;
+            assertThat(shadowConnectivityManager.getNetworkCallbacks().size(),
+                    is(equalTo(isAccountRegistered ? 1 : 0)));
+
+            if (!isAccountRegistered) {
+                return; // Without accounts, not network callbacks must be updates
```
muthenberg: type

PR #170, opened 2019-05-29
Task/stad 94 pausenerkennung

hb0:


persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
```diff
@@ -565,6 +565,50 @@ public void deletePoint3dData(final long measurementIdentifier) {
...
+                cleanedTrack.add(track);
+            }
+        }
+        Log.d(TAG,
+                String.format("Cleaned track: %d of %d sub tracks and %d of %d locations filtered.",
+                        (rawTracks.size() - cleanedTrack.size()), rawTracks.size(),
+                        (numberOfRawPoints - numberOfCleanPoints), numberOfRawPoints));
+        return cleanedTrack;
+    }
+
```
muthenberg: Doing such stuff in Java code is most likely a performance bottleneck. This could be achieved directly in SQL which should be way faster and avoid loading points we do not need into memory on the first hand. We should probably talk about this. I have a bad feeling about this (https://youtu.be/S74rvpc6W60?t=9).

persistence/src/main/java/de/cyface/persistence/DefaultLocationCleaningStrategy.java
```diff
@@ -91,19 +88,17 @@ public boolean isClean(@NonNull final GeoLocation location) {
...
-            Log.d(TAG, "GeoLocation with low speed filtered: " + location.getSpeed());
             return false;
         }
 
         // Ignore locations which are too far away from their previous location (upper speed limit)
+        // noinspection RedundantIfStatement - for readability
         if (location.getSpeed() > UPPER_SPEED_THRESHOLD) {
-            Log.d(TAG, "GeoLocation with high speed filtered: " + location.getSpeed());
             return false;
         }
```
muthenberg: You could just do `return location.getAccuracy() < UPPER_ACCURACY_THRESHOLD && location.getSpeed() > LOWER_SPEED_THRESHOLD && location.getSpeed() < UPPER_SPEED_THRESHOLD`
There actually is no need to be so verbose.

persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
hb0: You are totally right, I focused only on reusing existing code instead of minimizing resources. I'll add a custom cursor for this database access.

hb0: Implemented. Please review the two new commits and approve.

PR #173, opened 2019-06-05
Fix 4.1.1/stad 118

hb0:


.gitignore
```diff
@@ -13,3 +13,6 @@
 synchronization/src/main/res/raw/**/*.jks
 synchronization/src/main/res/raw/**/*.cer
 synchronization/src/main/res/raw/**/*.pem
+
+# git backup files
+*.orig
```
muthenberg: BTW.: https://stackoverflow.com/questions/1251681/git-mergetool-generates-unwanted-orig-files

PR #176, opened 2019-07-10
Bug/stad 140 cached locations before resume lead to slicing view error

hb0:


persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
```diff
@@ -656,18 +656,20 @@ private Cursor loadEvents(final long measurementIdentifier) {
...
-                    while (location.getTimestamp() < resumeEventTime) {
-                        geoLocationCursor.moveToNext();
-                        location = loadGeoLocation(geoLocationCursor);
-                    }
-                    geoLocationCursor.moveToPrevious();
-                    break; // Next track reached
+                // Ignore locations between pause and resume event (STAD-140)
+                while (location.getTimestamp() < resumeEventTime) {
+                    geoLocationCursor.moveToNext();
+                    location = loadGeoLocation(geoLocationCursor);
```
muthenberg: Why are we loading the Geo location here?

PR #177, opened 2019-07-17
Add API to load events from persistence STAD-143

hb0:


persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
```diff
@@ -489,7 +502,7 @@ public void delete(final long measurementIdentifier) {
      *
      * @param measurementIdentifier The {@code Measurement} id of the data to remove.
      */
-    public void deletePoint3dData(final long measurementIdentifier) {
```
hb0: This is no longer used by Movebis/CyfaceDCS. Thus, it can be private.

persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
```diff
@@ -534,7 +547,7 @@ public void deletePoint3dData(final long measurementIdentifier) {
         Cursor geoLocationCursor = null;
         Cursor eventCursor = null;
         try {
-            eventCursor = loadEvents(measurementIdentifier);
```
hb0: had to rename it or else it clashes with the

public List<Event> loadEvents(mid) method introduced. But the former method was anyway private.

PR #188, opened 2019-09-19
Task/vic 9 modal shifts

hb0:
These changes were necessary to implement the Modality Shift Functionality (including the view, delete and add Event functions to manage those shifts).


hb0:
wrong branch name

PR #190, opened 2019-09-24
Fix CusorIndexOutOfBoundsException STAD-171

hb0:


synchronization/src/androidTest/java/de/cyface/synchronization/TestUtils.java
```diff
@@ -25,14 +25,14 @@
...
      * A password used by the tests to set up a Cyface account for synchronization.
      */
-    final static String DEFAULT_PASSWORD = "secret";
+    final static String DEFAULT_PASSWORD = "7vXnwVAXfWXKbC32";
     /**
      * Path to an API available for testing.
      */
     @SuppressWarnings("unused") // because this is used in the cyface flavour
-    final static String TEST_API_URL = "https://s1.cyface.de/api/v2";
+    final static String TEST_API_URL = "https://s2.cyface.de/api/v2";
```
muthenberg: s2 should not be used for testing

synchronization/src/androidTest/java/de/cyface/synchronization/TestUtils.java
```diff
@@ -25,14 +25,14 @@
     /**
      * An username used by the tests to set up a Cyface account for synchronization.
      */
-    final static String DEFAULT_USERNAME = "test";
+    final static String DEFAULT_USERNAME = "test@cyface.de";
     /**
      * A password used by the tests to set up a Cyface account for synchronization.
      */
-    final static String DEFAULT_PASSWORD = "secret";
+    final static String DEFAULT_PASSWORD = "7vXnwVAXfWXKbC32";
```
muthenberg: Please do not provide passwords on s2 on github. An please change that one now.

persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
```diff
@@ -702,14 +702,22 @@ private Cursor loadEventsCursor(final long measurementIdentifier) {
                 // Pause reached: Move geoLocationCursor to the first location of the next sub-track
                 // We do this to ignore locations between pause and resume event (STAD-140)
                 while (location.getTimestamp() < resumeEventTime) {
-                    geoLocationCursor.moveToNext();
+                    if (!geoLocationCursor.moveToNext()) {
```
muthenberg: You should probably use hasNext here, since the while loop is responsible for the moveToNext. 

Why is the if condition not part of the while loop condition? It gets called on each iteration at first, which is exactly what the while loop does, isn't it?

persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
```diff
@@ -702,14 +702,22 @@ private Cursor loadEventsCursor(final long measurementIdentifier) {
...
                 }
 
-                // We reached the first location of the next sub-track, thus, move pointer back one step
-                // As the next outer while iteration will move the pointer one step forward
-                geoLocationCursor.moveToPrevious();
+                // Check if there is another sub-track after this sub-tracks resume event
+                if (location.getTimestamp() >= resumeEventTime) {
+                    // We reached the first location of the next sub-track, thus, move pointer back one step
+                    // As the next outer while iteration will move the pointer one step forward
+                    geoLocationCursor.moveToPrevious();
```
muthenberg: this is a really ugly control flow. Isn't there some better way of doing this?

synchronization/src/androidTest/java/de/cyface/synchronization/TestUtils.java
hb0: The problem is that we have some (old) tests which still rely on an actual API although (from the code) they should use mocked code. In order to be able to fix the bug in the current releases I need to keep the CI happy.

Thus, I am using the s2 but with a special test account which we can clean up easily.

**Another option** would be to disable all API dependent tests until we have time to investigate this dependency (which we did already some time ago but without sustainable results).

synchronization/src/androidTest/java/de/cyface/synchronization/TestUtils.java
hb0: See above: It's a special account for integration testing only. Although it's a dirty workaround. Depending on weather I should disable the API dependent tests I can just remove this account then and delete the credentials here.

persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
hb0: Thought it's better readable but it's changed now.

hb0: There is no "hasNext" for cursors. Thus, moveToNext() is used. But see refactored code for an alternative implementation.

persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
hb0: Here is an alternative implementation but it's even worse to understand I think:
https://github.com/cyface-de/android-backend/compare/fix/4.2.3/STAD-171...tmp/alternateImplementation1?expand=1

Another alternative would be to
- first slice the tracks into sub tracks
- then remove the locations between the pause and resume of the sub tracks

Please provide feedback if this should be implemented before a put more time into this.

synchronization/src/androidTest/java/de/cyface/synchronization/TestUtils.java
muthenberg: Disable them and create a task.

PR #191, opened 2019-09-24
Fix/4.2.3/stad 171

hb0:
see other PR #190 - same changes only here are the merge commits form master (already verified) included.


hb0:
I'll rebase the dev on top of the new master separately.

PR #194, opened 2019-09-30
Fix/4.2.4/vic 78 index out of bounds exception

hb0:
Bug fix for the refactored loadTracks() method (which was done in the 4.2.3 PR review).
Added reproducing test for this bug.

This needs to be released asap for the SR app as this produces crashes.

### DIFF
--- a/persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
+++ b/persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
@@ -65,7 +65,7 @@ import de.cyface.utils.Validate;
  *
  * @author Klemens Muthmann
  * @author Armin Schnabel
- * @version 16.0.0
+ * @version 17.0.1
  * @since 2.0.0
  */
 public class PersistenceLayer<B extends PersistenceBehaviour> {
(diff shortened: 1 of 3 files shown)


persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
```diff
@@ -745,7 +745,13 @@ private Track collectNextSubTrack(@NonNull final Cursor geoLocationCursor, @NonN
         while (location.getTimestamp() <= pauseEventTime) {
```
muthenberg: I would rather add a track.add(location) in front of the while loop and put the !geoLocationCursor.moveToNext() into the loop condition. However I wonder whether the initial location may already be null? Is this possible?

persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
```diff
@@ -745,7 +745,13 @@ private Track collectNextSubTrack(@NonNull final Cursor geoLocationCursor, @NonN
...
             track.add(location);
-            geoLocationCursor.moveToNext();
+
+            // Stop if this is the last GeoLocation
+            if (!geoLocationCursor.moveToNext()) {
+                break;
+            }
+
+            // Load next GeoLocation to check it's timestamp in next iteration
             location = loadGeoLocation(geoLocationCursor);
```
muthenberg: You could also just let loadGeoLocation return null if there is no further location and finish the loop as soon as location becomes null. This would prevent the above mentioned problem.

PR #204, opened 2019-12-10
Task/dat 129/upgrade android sdk to 29

hb0:
The SDK now targets Android Q.
This allows us to declare that our SDK does not need the new "Background Location" permission.
This permission is only required for non-user-initiated location requests.
We, however, only check the location when the App is in foreground or in our foreground service (datacapturing).

However, when targeting Android Q we need the new "foregroundServiceType: location" flag in the manifest for the datacapturing service for this to work.

Thus, this flag was added.

### DIFF
--- a/datacapturing/src/main/AndroidManifest.xml
+++ b/datacapturing/src/main/AndroidManifest.xml
@@ -4,6 +4,7 @@
     <!-- Required to keep the app from being send to sleep by the system. -->
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <!-- Required to access GNSS location providers -->
+    <!-- (!) We do *not* need ACCESS_BACKGROUND_LOCATION as we only need location in Foreground Service-->
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <!-- Required to change periodic synchronisation intervals in the background -->
     <uses-permission android:name="android.permission.WRITE_SYNC_SETTINGS" />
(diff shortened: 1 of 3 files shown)


datacapturing/src/main/AndroidManifest.xml
```diff
@@ -15,11 +16,19 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
 
     <application>
-        <!-- Declares the data capturing service which is not exported to other applications. Most important is the android:process declaration with the ":" in the beginning. It tells the system to run the service in its own process. -->
```
hb0: As the auto-format is not configured to break this comment down I did this manually.

datacapturing/src/main/AndroidManifest.xml
```diff
@@ -15,11 +16,19 @@
...
+        <!-- Foreground services in Q+ require type 'location' to get the location when the app
+        is minimized. ACCESS_FINE_LOCATION and foregroundServiceType="location" is enough on Q+
+        as long as we target the Q+ API. -->
+        <!-- The "unused attribute" warning for foregroundServiceType can be ignored as this declaration
+        requirement was added in Android Q so no handling is required for lower APIs. -->
         <service
             android:name=".backend.DataCapturingBackgroundService"
             android:description="@string/capturing_service_description"
             android:exported="false"
+            android:foregroundServiceType="location"
```
hb0: This is the important thing in this commit

hb0:
Closing this PR as the branch name changed (moved to other project)

PR #208, opened 2020-01-08
Task/cy 5065 GitHub actions

hb0:


persistence/src/main/java/de/cyface/persistence/Constants.java
```diff
@@ -20,17 +16,15 @@
     /**
      * The file extension of the measurement file which is transmitted on synchronization.
      */
-    public static final String TRANSFER_FILE_EXTENSION = "cyf";
+    public static final String TRANSFER_FILE_EXTENSION = "ccyf";
```
hb0: I fixed the typo "cyf" instead of "ccyf" as transfer file name extension already previously but this was accidentally done the "hard-coded" way instead of using this constant which was not in use until now.

README.md
```diff
@@ -1,6 +1,10 @@
 Cyface Android SDK
 ========================
 
+[<img src="https://github.com/cyface-de/android-backend/workflows/Gradle%20Build/badge.svg">](https://github.com/cyface-de/android-backend/actions)
+[<img src="https://github.com/cyface-de/android-backend/workflows/Gradle%20Publish/badge.svg">](https://github.com/cyface-de/android-backend/actions)
+[![Build Status](https://app.bitrise.io/app/3f7da1f067e2d183/status.svg?token=Ggdr1abrQO7N_Pm1uI4kjg&branch=master)](https://app.bitrise.io/app/3f7da1f067e2d183)
```
hb0: This token is only for this "batch icon" - it's generated automatically by Bitrise for that purpose only. No harm can be done with it.

muthenberg: I think we should switch from Bitrise to Github Actions in the foreseeable future, like with all the other projects. As far as I noticed there are workflows for Android builds.

.github/workflows/gradle_build.yml
```diff
@@ -0,0 +1,41 @@
...
+          mkdir -p synchronization/src/main/res/raw
+          touch synchronization/src/main/res/raw/truststore.jks
+
+      - name: Create local.properties with a read token
+        run: |
+          # Use a personal read token to install the Cyface Utils package
+          echo "github.user=${{ secrets.GITHUB_READ_ACCOUNT }}" > local.properties
+          echo "github.token=${{ secrets.GITHUB_READ_TOKEN }}" >> local.properties
+      - name: Build with Gradle
+        run: ./gradlew build
```
muthenberg: Is it possible to run the tests here as well?

README.md
hb0: agree

.github/workflows/gradle_build.yml
hb0: The UnitTests are already executed by the "Build with Gradle" step (https://github.com/cyface-de/android-backend/commit/1fea91627bfcf1c9814bd454332d0dac05de302f/checks?check_suite_id=389946445).

The Integration test require an emulator on Github Actions.
I created a Ticket for checking out if this is (easily) possible: 
https://cyface.atlassian.net/browse/CY-5074

PR #215, opened 2020-02-27
Task/mov 945 fix exception after multipart code refactoring

hb0:


synchronization/src/main/java/de/cyface/synchronization/HttpConnection.java
```diff
@@ -214,7 +214,7 @@ public HttpResponse post(@NonNull final HttpURLConnection connection, @NonNull f
 
                 // Write MultiPart file parts
                 for (FilePart filePart : fileParts) {
-                    filePart.writeTo(outputStream, progressListener);
+                    bytesWrittenToOutputStream += filePart.writeTo(outputStream, progressListener);
```
hb0: This was the actual bug which produced the exception in this ticket

synchronization/src/androidTestMovebis/java/de/cyface/synchronization/SyncPerformerTest.java
```diff
@@ -262,7 +262,8 @@ public void testSendData_toActualApi() throws CursorIsNullException, NoSuchMeasu
             final GeoLocation startLocation = tracks.get(0).getGeoLocations().get(0);
             final List<GeoLocation> lastTrack = tracks.get(tracks.size() - 1).getGeoLocations();
             final GeoLocation endLocation = lastTrack.get(lastTrack.size() - 1);
-            final SyncAdapter.MetaData metaData = new SyncAdapter.MetaData(startLocation, endLocation, "testDeviceId",
+            final String deviceId = "testDevi-ce00-42b6-a840-1b70d30094b8"; // Must be a valid UUID
+            final SyncAdapter.MetaData metaData = new SyncAdapter.MetaData(startLocation, endLocation, deviceId,
```
hb0: This was another bug I found while executing this bug-reproducing test.
This test is test-API dependent, thus, the CI did not catch those bugs.

PR #221, opened 2020-06-24
Bug/dat 382 event time offset set more than once

hb0:
While checking the code to see if we could synchronize the sensor-data of multiple devices by synchronizing the System clocks I noticed, that the `sensorTimeOffset` could theoretically be reset and lead to time-shifts in the data if:
* the first `sensorTimeOffset` calculation is `0` ms
* then the code would recalculate the offset until the offset is `!= 0`

This is why I'm using `Long` instead of `long` for the offset.
I say "theoretically" because the delay between the `event` and the `onSensorChanged` call was between 2 and 9 ms on our reference device. Thus, it might never be `0 ms` on real devices.

----------

In order to synchronize the sensor data we need to make sure the delay between the first `event` and it's `onSensorChanged(event)` call does not influence the `eventTimeOffsetMillis` and, thus, shift the sensor data a few milliseconds depending on the device.

To do this we're now only using the `SystemClock` to calculate the `eventTimeOffsetMillis` in the documented case where the `event.time` is the `elapsedRealTimeNanos` (which is the time since boot including the (deep) sleep time).

For the few devices where the `event.time` is the `currentTimeNanos` we're using an offset of `0` which should explain itself.

### DIFF
--- a/datacapturing/src/main/java/de/cyface/datacapturing/PongReceiver.java
+++ b/datacapturing/src/main/java/de/cyface/datacapturing/PongReceiver.java
@@ -28,7 +28,6 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
-import android.os.Build;
 import android.os.Handler;
 import android.os.HandlerThread;
 import android.os.Process;
@@ -203,11 +202,7 @@ public class PongReceiver extends BroadcastReceiver {
 
     private void quitBackgroundHandlerThread() {
         if (Thread.currentThread().getName().equals(pongReceiverThread.getName())) {
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR2) {
-                pongReceiverThread.quitSafely();
-            } else {
-                pongReceiverThread.quit();
-            }
+            pongReceiverThread.quitSafely();
         }
     }
 }
--- a/datacapturing/src/main/java/de/cyface/datacapturing/PongReceiver.java
+++ b/datacapturing/src/main/java/de/cyface/datacapturing/PongReceiver.java
@@ -28,6 +28,7 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
+import android.os.Build;
 import android.os.Handler;
 import android.os.HandlerThread;
 import android.os.Process;
@@ -202,7 +203,11 @@ public class PongReceiver extends BroadcastReceiver {
 
     private void quitBackgroundHandlerThread() {
         if (Thread.currentThread().getName().equals(pongReceiverThread.getName())) {
-            pongReceiverThread.quitSafely();
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR2) {
+                pongReceiverThread.quitSafely();
+            } else {
+                pongReceiverThread.quit();
+            }
         }
     }
 }
--- a/datacapturing/src/main/java/de/cyface/datacapturing/backend/CapturingProcess.java
+++ b/datacapturing/src/main/java/de/cyface/datacapturing/backend/CapturingProcess.java
@@ -244,7 +244,7 @@ public abstract class CapturingProcess implements SensorEventListener, LocationL
     @Override
     public synchronized void onSensorChanged(final @NonNull SensorEvent event) {
         if (eventTimeOffsetMillis == null) {
-            eventTimeOffsetMillis = eventTimeOffset(event);
+            eventTimeOffsetMillis = eventTimeOffset(event.timestamp);
         }
         Validate.notNull(eventTimeOffsetMillis);
         long thisSensorEventTime = event.timestamp / 1_000_000L + eventTimeOffsetMillis;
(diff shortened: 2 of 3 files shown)


datacapturing/src/main/java/de/cyface/datacapturing/backend/CapturingProcess.java
```diff
@@ -284,6 +279,64 @@ public synchronized void onSensorChanged(final @NonNull SensorEvent event) {
...
+                    + "to zero ms");
+            return 0;
+        }
+        // Possible case, but unknown if actually used by manufacturers (upTime)
+        // If we calculate a static offset between currentTime and upTime this would lead to time
+        // shifts when the device is in (deep) sleep again. We would need to calculate the offset
+        // dynamically for each event which is quite heavy. Thus, we throw an exception to see
+        // in the Play Store if this actually happens on devices. If so, this could be tested using:
+        // https://developer.android.com/training/monitoring-device-state/doze-standby#testing_doze_and_app_standby
+        throw new IllegalStateException("The event.time seems to be upTime. In this case we cannot"
```
muthenberg, line 336: Is this wise? We might get a plethora of crashes from the field, if we introduce this here.

datacapturing/src/main/java/de/cyface/datacapturing/backend/CapturingProcess.java
```diff
@@ -265,9 +262,7 @@ public synchronized void onSensorChanged(final @NonNull SensorEvent event) {
                 rotations.clear();
                 directions.clear();
                 lastNoGeoLocationFixUpdateTime = thisSensorEventTime;
-            } catch (SecurityException e) {
-                throw new IllegalStateException(e);
-            } catch (DataCapturingException e) {
+            } catch (SecurityException | DataCapturingException e) {
```
muthenberg, line 265: Probably should move changes for updates due to newer Android version to another PR.

datacapturing/src/main/java/de/cyface/datacapturing/PongReceiver.java
```diff
@@ -203,11 +202,7 @@ public void onReceive(final @NonNull Context context, final @NonNull Intent inte
 
     private void quitBackgroundHandlerThread() {
         if (Thread.currentThread().getName().equals(pongReceiverThread.getName())) {
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR2) {
-                pongReceiverThread.quitSafely();
-            } else {
-                pongReceiverThread.quit();
-            }
+            pongReceiverThread.quitSafely();
```
muthenberg, line 205: Probably should move changes due to Android Update to another PR.

datacapturing/src/main/java/de/cyface/datacapturing/backend/CapturingProcess.java
```diff
@@ -243,14 +243,11 @@ public void onProviderDisabled(final String provider) {
...
-        // been the other way around.
-        if (eventTimeOffset == 0) {
-            // event.timestamp on Nexus5 with Android 6.0.1 seems to be the uptime in Nanoseconds (resets with
-            // rebooting)
-            eventTimeOffset = System.currentTimeMillis() - event.timestamp / 1_000_000L;
+        if (eventTimeOffsetMillis == null) {
+            eventTimeOffsetMillis = eventTimeOffset(event.timestamp);
         }
-        long thisSensorEventTime = event.timestamp / 1_000_000L + eventTimeOffset;
+        Validate.notNull(eventTimeOffsetMillis);
```
muthenberg, line 249: since eventTimeOffset() returns long this check will never be reached. If eventTimeOffset() would return null there would be an NPE inside the if statement, right?

datacapturing/src/test/java/de/cyface/datacapturing/backend/CapturingProcessTest.java
```diff
@@ -141,6 +145,30 @@ public void testCaptureSensorDataAlongWithGeoLocation() {
         assertThat(testListener.getCapturedLocations(), Matchers.hasSize(2));
     }
 
+    /**
+     * Tests that the correct `eventTimeOffset` is calculated for known `event.time` implementations.
+     */
+    @Test
+    public void testEventTimeOffset() {
```
muthenberg, line 152: *thumbsup* for adding a test case. :D

datacapturing/src/main/java/de/cyface/datacapturing/backend/CapturingProcess.java
hb0: I'm not sure what's better:
1. only release this change to small test groups at the beginning to see if this exception actually exists
2. handle this softly and never know if this actually happens. Soft handling would mean (A) static offset --> data might be shifted multiple times or (B) re-calculate offset on each sensor update (or a selection of those), without knowing if this is necessary or (C) ...

I would go (and went for): 1.
As long as we introduce this slowly (own test phones, research partner phones, public test app users) we actually get a good sense if this is a exception which can occur or not and can eventually release this change to paid customer's users.

datacapturing/src/main/java/de/cyface/datacapturing/backend/CapturingProcess.java
hb0: right.

PR #238, opened 2022-03-16
Task/dat 965 support registration

hb0:
Adds the support for:
- 428: Account not activated response from the Auth server
- unexpected error codes (before, the processed crashed, e.g. before the 428 handling was added)


muthenberg (approved):
Looking good

PR #245, opened 2022-06-02
Upgrade transfer file and persistence file version to 3

hb0:
Makes the SDK use the new `serializer` 2.0.0 with binary format 3


build.gradle
```diff
@@ -21,7 +21,7 @@
  *
  * @author Armin Schnabel
  * @author Klemens Muthmann
- * @version 2.8.2
+ * @version 2.8.3
```
hb0, line 24: TODO: upgrade serializer dependency

PR #251, opened 2022-11-08
[RFR-104] Fix bug in location serialization

hb0:
This PR fixes the bug where the location values where doubled when more than 10.000 locations are captured. The bug was captured by the deserialization pipeline as the locations then had a latitute > 90 degrees:

```
timestamp, lat, lon, speed, etc.
1666003984000,49.604472,11.366676,27.25,1.56,CAR,294.5537253607995,294.5537253607995,10003000,10003000
3332007969000,99.209165,22.733509,54.52,3.1,CAR,294.5537253607995,294.5537253607995,10003000,10003000
```

The sensor data was not affected, only the location data.

I added a bug reproducing test before I fixed it.


hb0 (commented):
pre-reviewed

muthenberg (approved):

PR #254, opened 2023-01-06
Task/rfr 253 replace deprecated handler constructor

hb0:
This PR replaces the deprecated `Handler()` constructor with the constructor which injects an explicit Looper.

Unfortunately, this does not fix the `rebind()` bug, but I might be able to trace down the cause of the `rebind` bug with the reproducer app, but that's another story.

### DIFF
--- a/datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
+++ b/datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
@@ -56,7 +56,6 @@ import android.net.ConnectivityManager;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.IBinder;
-import android.os.Looper;
 import android.os.Message;
 import android.os.Messenger;
 import android.os.RemoteException;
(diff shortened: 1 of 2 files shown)


hb0 (commented):
pre-reviewed

datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
```diff
@@ -278,7 +279,7 @@ public DataCapturingService(@NonNull final Context context, @NonNull final Strin
             throw new SetupException("Android connectivity manager is not available!");
         }
         surveyor = new WiFiSurveyor(context, connectivityManager, authority, accountType);
-        fromServiceMessageHandler = new FromServiceMessageHandler(context, this);
+        fromServiceMessageHandler = new FromServiceMessageHandler(context.getMainLooper(), context, this);
```
muthenberg, line 282: You are getting the looper from the context. So why do you not just call `context.getMainLooper()` inside the super call? That way you would not need to change this line.

muthenberg (changes requested):

datacapturing/src/main/java/de/cyface/datacapturing/DataCapturingService.java
hb0: Good point, I reverted this API change.

PR #255, opened 2023-01-08
Task/stad 363 average speed

hb0:
Will be part of #257

### DIFF
--- a/persistence/build.gradle
+++ b/persistence/build.gradle
@@ -45,6 +45,10 @@ android {
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
         }
     }
+
+    lintOptions {
+        abortOnError false
+    }
 }
 
 dependencies {
--- a/persistence/src/main/java/de/cyface/persistence/DatabaseHelper.java
+++ b/persistence/src/main/java/de/cyface/persistence/DatabaseHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Cyface GmbH
+ * Copyright 2017-2023 Cyface GmbH
  *
  * This file is part of the Cyface SDK for Android.
  *
--- a/persistence/src/main/java/de/cyface/persistence/DatabaseHelper.java
+++ b/persistence/src/main/java/de/cyface/persistence/DatabaseHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2023 Cyface GmbH
+ * Copyright 2017 Cyface GmbH
  *
  * This file is part of the Cyface SDK for Android.
  *
@@ -58,11 +58,8 @@ class DatabaseHelper extends SQLiteOpenHelper {
     /**
      * Increase the DATABASE_VERSION if the database structure changes with a new update
      * but don't forget to adjust onCreate and onUpgrade accordingly for the new structure and incremental upgrade
-     * <p>
-     * <b>Attention:</b> As this is a fork of the `main`, another version range is selected as the
-     * main branch database uses another database migration "route". Their ancestor is version 16.
      */
-    private final static int DATABASE_VERSION = 1_000_017; // FIXME: Blocked by [STAD-380]
+    private final static int DATABASE_VERSION = 16;
     /**
      * The table containing all the measurements, without the corresponding data. Data is stored in one table per type.
      */
@@ -156,10 +153,6 @@ class DatabaseHelper extends SQLiteOpenHelper {
             geoLocationsTable.onUpgrade(database, fromVersion, toVersion);
             identifierTable.onUpgrade(database, fromVersion, toVersion);
             eventTable.onUpgrade(database, fromVersion, toVersion);
-
-            // This is a fork of the main branch, so we use another version range. [STAD-380]
-            // This avoids 1 mio log entries "Upgrading database from version ... to ..."
-            fromVersion = fromVersion == 16 ? 1_000_016 : fromVersion;
         }
     }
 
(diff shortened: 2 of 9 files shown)


persistence/src/main/java/de/cyface/persistence/DatabaseHelper.java
```diff
@@ -58,8 +58,11 @@ class DatabaseHelper extends SQLiteOpenHelper {
     /**
      * Increase the DATABASE_VERSION if the database structure changes with a new update
      * but don't forget to adjust onCreate and onUpgrade accordingly for the new structure and incremental upgrade
+     * <p>
+     * <b>Attention:</b> As this is a fork of the `main`, another version range is selected as the
+     * main branch database uses another database migration "route". Their ancestor is version 16.
      */
-    private final static int DATABASE_VERSION = 16;
+    private final static int DATABASE_VERSION = 1_000_017; // FIXME: Blocked by [STAD-380]
```
hb0, line 65: see comment in the linked ticket

persistence/src/main/java/de/cyface/persistence/DatabaseHelper.java
```diff
@@ -153,6 +156,10 @@ public void onUpgrade(final SQLiteDatabase database, final int oldVersion, final
             geoLocationsTable.onUpgrade(database, fromVersion, toVersion);
             identifierTable.onUpgrade(database, fromVersion, toVersion);
             eventTable.onUpgrade(database, fromVersion, toVersion);
+
+            // This is a fork of the main branch, so we use another version range. [STAD-380]
+            // This avoids 1 mio log entries "Upgrading database from version ... to ..."
+            fromVersion = fromVersion == 16 ? 1_000_016 : fromVersion;
```
hb0, line 162: see comment above

persistence/src/main/java/de/cyface/persistence/MeasurementTable.java
```diff
@@ -140,14 +152,106 @@ public void onUpgrade(final SQLiteDatabase database, final int fromVersion, fina
...
+                break;
         }
 
     }
 
+    /**
+     * Adds columns related to average speed to table and calculates the values for existing data.
+     * <p>
+     * Like for newly captured measurements, the average speed is only based on locations not filtered
+     * by the {@link DefaultLocationCleaningStrategy}.
```
hb0, line 174: I think it's expected, that the average speed does not decrease when the user is not moving for a short time. (also, because the customer said "should be like distance calculation)
Thus, I'm using the location cleaning strategy here in the data migration and in the actual data capturing before calculating average speed, to filter "invalid" locations.

persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
```diff
@@ -1023,6 +1028,25 @@ public void setDistance(final long measurementIdentifier, final double newDistan
...
+     * @throws NoSuchMeasurementException if there was no {@code Measurement} with the id
+     *             {@param measurementIdentifier}.
+     */
+    public void setAverageSpeed(final long measurementIdentifier, final double newSpeedSum, final int newSpeedSamples)
+            throws NoSuchMeasurementException {
+        final ContentValues values = new ContentValues();
+        values.put(COLUMN_SPEED_SUM, newSpeedSum);
+        values.put(COLUMN_SPEED_COUNTER, newSpeedSamples);
+        updateMeasurement(measurementIdentifier, values);
+    }
```
hb0, line 1048: My understanding of the feature request was:
- the average speed per measurement should be calculated
- i.e. when we pause the first track of the measurement and resume, the speed samples of all sub-tracks are used to calculate the average speed
- for this to work, we cannot save "averageSpeed" to the database but have to save the speed sum and counter.

persistence/src/test/java/de/cyface/persistence/DatabaseHelperTest.java
```diff
@@ -127,10 +128,73 @@ public void tearDown() {
         db.close();
     }
 
+    /**
+     * Test upgrading the {@link MeasurementTable} to Database V1000017.
+     * <p>
+     * We test the speed fields are calculated correctly, with and without locations.
+     */
+    @Test
+    public void testMigrationV16ToV1000017() {
```
hb0, line 137: version "1_000_017" see comments above

persistence/src/test/java/de/cyface/persistence/DatabaseHelperTest.java
```diff
@@ -127,10 +128,73 @@ public void tearDown() {
         db.close();
     }
 
+    /**
+     * Test upgrading the {@link MeasurementTable} to Database V1000017.
+     * <p>
+     * We test the speed fields are calculated correctly, with and without locations.
+     */
```
hb0, line 135: I've also tested the database migration with our test-app

hb0 (commented):
pre-reviewed

persistence/build.gradle
```diff
@@ -45,6 +45,10 @@ android {
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
         }
     }
+
+    lintOptions {
+        abortOnError false
+    }
```
hb0, line 51: Without this, we would need to fix ~ 30 lint errors here:

```
ret += deleteDataForMeasurement(database, selectedMeasurementsCursor
                                        .getLong(selectedMeasurementsCursor.getColumnIndex(BaseColumns._ID)));
```

=> getColumnIndex can be -1 when the column does not exist, but `cursor.getLong()` only accepts an index int > 0

I'm not sure if we want to do much "cleanup" work in the SDK 6 fork, so I added the "ignore lint" option instead.

PR #257, opened 2023-01-09
Task/stad 367 measurement duration

hb0:
This PR adds:
- average speed calculation
- measurement duration calculation

**Recommendation:**
Review PR via "Files changed" or else you'll review much more code then necessary (average speed stored in database which was reverted).

### DIFF
--- a/persistence/build.gradle
+++ b/persistence/build.gradle
@@ -45,6 +45,10 @@ android {
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
         }
     }
+
+    lintOptions {
+        abortOnError false
+    }
 }
 
 dependencies {
--- a/persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
+++ b/persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
@@ -25,6 +25,10 @@ import static de.cyface.persistence.MeasurementTable.COLUMN_MODALITY;
 import static de.cyface.persistence.MeasurementTable.COLUMN_PERSISTENCE_FILE_FORMAT_VERSION;
 import static de.cyface.persistence.MeasurementTable.COLUMN_STATUS;
 import static de.cyface.persistence.MeasurementTable.COLUMN_TIMESTAMP;
+import static de.cyface.persistence.model.Event.EventType.LIFECYCLE_PAUSE;
+import static de.cyface.persistence.model.Event.EventType.LIFECYCLE_RESUME;
+import static de.cyface.persistence.model.Event.EventType.LIFECYCLE_START;
+import static de.cyface.persistence.model.Event.EventType.LIFECYCLE_STOP;
 import static de.cyface.persistence.model.MeasurementStatus.FINISHED;
 import static de.cyface.persistence.model.MeasurementStatus.OPEN;
 import static de.cyface.persistence.model.MeasurementStatus.SYNCED;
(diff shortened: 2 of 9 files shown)


persistence/build.gradle
```diff
@@ -45,6 +45,10 @@ android {
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
         }
     }
+
+    lintOptions {
+        abortOnError false
+    }
```
hb0, line 51: Without this, we would need to fix ~ 30 lint errors here:

```
ret += deleteDataForMeasurement(database, selectedMeasurementsCursor
                                        .getLong(selectedMeasurementsCursor.getColumnIndex(BaseColumns._ID)));
```

=> getColumnIndex can be -1 when the column does not exist, but `cursor.getLong()` only accepts an index int > 0

I'm not sure if we want to do much "cleanup" work in the SDK 6 fork, so I added the "ignore lint" option instead.

hb0 (commented):
pre-reviewed

persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
```diff
@@ -561,6 +565,96 @@ private void deletePoint3dData(final long measurementIdentifier) {
...
+            @NonNull final LocationCleaningStrategy locationCleaningStrategy) throws CursorIsNullException {
+
+        final List<Track> tracks = loadTracks(measurementIdentifier);
+        double speedSum = 0.0;
+        int speedCounter = 0;
+        for (final Track track : tracks) {
+            double sum = 0.0;
+            int counter = 0;
+            for (final GeoLocation location : track.getGeoLocations()) {
+                if (locationCleaningStrategy.isClean(location)) {
```
hb0, line 589: I think it's expected, that the average speed does not decrease when the user is not moving for a short time. (also, because the customer said "should be like distance calculation)
Thus, I'm adding speed samples only after the location cleaning strategy accepts the location captured.

My understanding of the feature request was:

the average speed per measurement should be calculated
i.e. when we pause the first track of the measurement and resume, the speed samples of all sub-tracks are used to calculate the average speed
for this to work, we cannot just save "averageSpeed" but have to save the speed sum and counter.

hb0 (commented):
pre-reviewed

persistence/build.gradle
muthenberg: Yeah. Just do not merge that to the main branch in the future.

persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
muthenberg: The customer did not say anything about "distance calculation". That is what I said. We should probably have a look into how other apps do this, before making a final decision.

muthenberg (changes requested):
We need to have a look into another Bike Statistics App like Komoot, Strava, Runtastic or the like, before deciding on whether to include pauses while averaging. I am not sure how this should be implemented.

persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
hb0: see discussion in Ticket

muthenberg: Seems there is no easy answer to this. Lets got with this implementation then.

muthenberg (approved):

PR #258, opened 2023-01-11
Task/stad 371 elevation profile

hb0:
This PR adds the features:
- capture GNSS altitude and vertical accuracy in Room database if available
- capture pressure in Room database if barometer is available
- calculate the ascend meters from barometer data is available or else from GNSS altitude if available. `null` is returned for measurements where no such data is available.

I suggest to review this this PR via `Files changed` as I first tried to persist the new data via SQLite API but then replaced this within a few commits with the Room API.
But commit-wise review is also ok-ish when you keep this in mind (skip attempt to store data in a second SQLite database, authority, content resolver).

Everything which is still missing is commented with TO-DO and I've sub-tasks for that, but you can already review the current state as it won't change a lot, I only need to:
- add tests with a track consisting of sub-tracks
- follow the migration steps required for `targetVersion` upgrade from `30` to `31` which is required to use `Room`


hb0 (commented):
Pre-viewed part 1

persistence/src/main/java/de/cyface/persistence/model/GeoLocationV6.java
```diff
@@ -0,0 +1,320 @@
...
+                && Double.compare(that.lon, lon) == 0 && Double.compare(that.speed, speed) == 0
+                && Double.compare(that.accuracy, accuracy) == 0 && measurementId == that.measurementId
+                && Objects.equals(altitude, that.altitude) && Objects.equals(verticalAccuracy, that.verticalAccuracy)
+                && Objects.equals(isValid, that.isValid);
+    }
+
+    @Override
+    public int hashCode() { // FIXME: w/ or w/o uid, measurement_fk? (definitely w/o isValid)
+        return Objects.hash(lat, lon, altitude, speed, accuracy, verticalAccuracy);
+    }
```
hb0, line 319: In `GeoLocation` we hashed:
- timestamp
- lat, lon, speed, accuracy
- isValid
The old class has no `measurementId` or database `identifier`, that data is only stored in the database, and in the case of `measurementId` only in the database and in `Measurement` and `Track` class, not `GeoLocation` directly.

Here in `GeoLocationV6 extends DataPointV6` we also have database `uid` and `measurementId`, but only after the data was loaded from the database, not while it's cached during capturing.

So the question would be: what to use for hashing?
- `DataPointV6.hash()` uses `uid` right now, but as `uid` is not set at cache-time I already questioned a few comments above if we should use `timestamp` for hashing, too, or maybe only `timestamp` as DataPoint(time=1L, mId=null) and `DataPoint(time=1L, mid=123L)` could considered the same object, just with and without mid set.

- `GeoLocationV6.hash()` should maybe use: `super.hash()` and `lat, lon, altitude, speed, accuracy, verticalAccuracy`, probably not `measurementId` as this is only set later but does not change the object "meant" by that object instance.

persistence/src/main/java/de/cyface/persistence/model/TrackV6.java
```diff
@@ -0,0 +1,120 @@
...
+ * <p>
+ * A {@code TrackV6} begins with the first data point of each type collected after start or resume was triggered
+ * and stops with the last collected data point of each type before the next resume command is triggered or when the
+ * very last location is reached.
+ *
+ * @author Armin Schnabel
+ * @version 1.0.0
+ * @since 6.3.0
+ */
+public class TrackV6 {
```
hb0, line 39: As `Track` only support `GeoLocation` I added this class which supports `GeoLocationV6` and `Pressure`.

This allowed me to use simpler code in `PersistenceLayer.loadAscend()` from pressure and location.altitude.

persistence/src/test/java/de/cyface/persistence/PersistenceLayerTest.java
```diff
@@ -0,0 +1,152 @@
...
+     * Based on the formula from {@code android.hardware.SensorManager#getAltitude(float, float)}.
+     *
+     * @param altitude The altitude to calculate the pressure for in meters above sea level.
+     * @param p0 The pressure at sea level for the specific weather condition.
+     * @return The atmospheric pressure in hPa.
+     */
+    @SuppressWarnings("SameParameterValue")
+    private float pressure(double altitude, float p0) {
+        return p0 * (float)Math.pow(1.0f - altitude / 44330.0f, 5.255f);
+    }
```
hb0, line 151: I resolved the formula which is used in `SensorManager.getAltitude(p, p0)` to get the pressure data I need to generate to test the pressure for specific border altitudes we support:
- 10_000 meters above sea level which is a bit more then the highest point on earth (mount everest)
- -500 meters below sea level which is a bit less than the lowest point on earth (death see).

[We don't support diving, flying and entering deep caves right now, which is fine I guess :)]

hb0: On mount everest the pressure is roughly 265 hPa (with slight variations because of the condition of the atmosphere (weather).

persistence/src/main/java/de/cyface/persistence/model/Pressure.java
```diff
@@ -0,0 +1,162 @@
...
+     */
+    public Pressure(final long timestamp, final double pressure) {
+        super(timestamp);
+
+        // Lowest/highest pressure on earth with a bounding box because of inaccuracy and weather. We only support
+        // measuring between death see and mt. everest, no flying, diving and caves are supported.
+        if (pressure < 250. || pressure > 1_100.) {
+            throw new IllegalArgumentException(String.format(Locale.US,
+                    "Illegal value for pressure. Is required to be between 250.0 and 1_100.0 but was %f.", pressure));
+        }
```
hb0, line 70: I checked the pressure at mt. everest and death see and checked the recorded pressure records on earth (while not flying, diving, entering deep caves) in history and added a bounding box because of weather conditions and sensor inaccuracy.

synchronization/src/main/AndroidManifest.xml
```diff
@@ -35,7 +35,8 @@
                     android:name="android.content.SyncAdapter"
                     android:resource="@xml/sync_adapter" />
         </service>
-        <service android:name="de.cyface.synchronization.AuthenticatorService">
+        <service android:name="de.cyface.synchronization.AuthenticatorService"
+          android:exported="false">
```
hb0, line 39: required in `targetVersion=31`

hb0 (commented):
Fully pre-reviewed

datacapturing/src/main/java/de/cyface/datacapturing/backend/CapturingProcess.java
```diff
@@ -169,11 +178,13 @@ public abstract class CapturingProcess implements SensorEventListener, LocationL
...
+        Sensor barometer = sensorService.getDefaultSensor(Sensor.TYPE_PRESSURE);
         sensorEventHandlerThread.start();
         Handler sensorEventHandler = new Handler(sensorEventHandlerThread.getLooper());
-        registerSensor(accelerometer, sensorEventHandler);
-        registerSensor(gyroscope, sensorEventHandler);
-        registerSensor(magnetometer, sensorEventHandler);
+        registerSensor(accelerometer, sensorEventHandler, delayBetweenSensorEventsInMicroseconds);
+        registerSensor(gyroscope, sensorEventHandler, delayBetweenSensorEventsInMicroseconds);
+        registerSensor(magnetometer, sensorEventHandler, delayBetweenSensorEventsInMicroseconds);
+        registerSensor(barometer, sensorEventHandler, 1_000_000);
```
muthenberg, line 187: If you do it like this, there is no need to store the delay in a property any longer, I guess.

persistence/src/main/java/de/cyface/persistence/dao/GeoLocationDao.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Insert
+    void insertAll(GeoLocationV6... locations);
+
+    @Delete
+    void delete(GeoLocationV6 location);
+
+    @SuppressWarnings("UnusedReturnValue")
+    @Query("DELETE FROM pressure WHERE measurement_fk = :measurementId")
+    int deleteItemByMeasurementId(long measurementId);
```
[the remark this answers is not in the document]
muthenberg: I usually would advise to keep it as simple as possible and remove all the clutter code.

persistence/src/main/java/de/cyface/persistence/dao/PressureDao.java
```diff
@@ -0,0 +1,58 @@
...
+ * @since 6.3.0
+ */
+@Dao
+public interface PressureDao {
+
+    @Query("SELECT * FROM pressure")
+    List<Pressure> getAll();
+
+    @Query("SELECT * FROM pressure WHERE uid IN (:pressureIds)")
+    List<Pressure> loadAllByIds(int[] pressureIds);
```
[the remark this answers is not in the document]
muthenberg: See my answer there as well. ;)

persistence/src/main/java/de/cyface/persistence/model/DataPoint.java
```diff
@@ -53,6 +53,7 @@ public long getTimestamp() {
         return timestamp;
     }
 
+    // FIXME: Should also check timestamp
```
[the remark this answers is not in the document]
muthenberg: According to the documentation, the identifier is system wide unique, so it should be enough to compare identifiers.

persistence/src/main/java/de/cyface/persistence/model/DataPointV6.java
```diff
@@ -0,0 +1,132 @@
...
+        if (o == null || getClass() != o.getClass())
+            return false;
+        DataPointV6 that = (DataPointV6)o;
+        return uid == that.uid && timestamp == that.timestamp;
+    }
+
+    @Override // FIXME: w/ or w/o uid, timestamp?
+    public int hashCode() {
+        return Objects.hash(uid);
+    }
```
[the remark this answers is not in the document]
muthenberg: I guess the cleanest solution would be to have two separate classes here (maybe a common interface). But I have no full overview on where this is used and what implications such a change has. I would leave it as is, if it works at the moment. Such changes should only be considered on the current HEAD. Only bugfixes and requested features are necessary on this branch.

persistence/src/main/java/de/cyface/persistence/model/GeoLocationV6.java
```diff
@@ -0,0 +1,320 @@
...
+    protected GeoLocationV6(final @NonNull Parcel in) {
+        super(in);
+        lat = in.readDouble();
+        lon = in.readDouble();
+        altitude = in.readDouble();
+        speed = in.readDouble();
+        accuracy = in.readDouble();
+        verticalAccuracy = in.readDouble();
+        isValid = in.readByte() != 0;
+        measurementId = in.readLong(); // FIXME: Added, do we need this? GeoLocation has no such field
```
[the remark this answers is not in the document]
muthenberg: Probably should include the id in the parcel, as that would be the expected behaviour. It might be always `null` for now, but that might change and it will be a hard to find bug, why the identifier is missing from the deserialized instances.

persistence/src/main/java/de/cyface/persistence/model/GeoLocationV6.java
muthenberg: I dimmly remember having put some thoughts into this years ago and not finding a satisfying architectural solution. Today I would probably consider to provide multiple classes or factor the changing parts out into their own class. But I think this is too much for this branch. As long as it works, a solution should only be added to the HEAD of main.

persistence/src/main/java/de/cyface/persistence/model/TrackV6.java
muthenberg: Different solution would probably be to add an interface for GeoLocation and use that with Track. But as it should not be a long lasting branch, this is probably fine.

muthenberg (changes requested):

persistence/schemas/de.cyface.persistence.DatabaseV6/1.json
```diff
@@ -0,0 +1,120 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 1,
+    "identityHash": "c3ec4780488bc825ddc6571e20a59607",
+    "entities": [
+      {
```
hb0, line 7: I added the database schema generated by Room to the version control to ensure the schema is as expected.

persistence/schemas/de.cyface.persistence.DatabaseV6/1.json
```diff
@@ -0,0 +1,120 @@
...
+        "tableName": "Pressure",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`pressure` REAL NOT NULL, `measurement_fk` INTEGER NOT NULL, `uid` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `timestamp` INTEGER NOT NULL)",
```
hb0, line 9: In order for `measurement_fk` to be `NOT NULL` as expected in the database I was not able to make the measurement id optional, as I did before.

For that reason I added a separate class for a persisted `Pressure` and `GeoLocationV6`.

This also cleans up the code as
- we only implement parcelable for the cached classes `Pressure` and `GeoLocationV6` but not for the persisted classes as we never send them though parcels
- we ensure that pressure data and locations cannot be inserted into the database without a linked measurement, as expected.

persistence/schemas/de.cyface.persistence.DatabaseV6/1.json
```diff
@@ -0,0 +1,120 @@
...
+            "uid"
+          ],
+          "autoGenerate": true
+        },
+        "indices": [],
+        "foreignKeys": []
+      },
+      {
+        "tableName": "Location",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`lat` REAL NOT NULL, `lon` REAL NOT NULL, `altitude` REAL, `speed` REAL NOT NULL, `accuracy` REAL NOT NULL, `vertical_accuracy` REAL, `measurement_fk` INTEGER NOT NULL, `uid` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `timestamp` INTEGER NOT NULL)",
```
hb0, line 47: see above

datacapturing/src/main/java/de/cyface/datacapturing/persistence/CapturingPersistenceBehaviour.java
```diff
@@ -122,9 +122,17 @@ public void storeData(final @NonNull CapturedData data, final long measurementId
...
-            final Pressure pressure = pressures.get(pressures.size() - 1);
+            // Calculating the average pressure to be less dependent on random outliers
+            double sum = 0.;
+            for (Pressure p : pressures) {
+                sum += p.getPressure();
+            }
+            final double averagePressure = sum / pressures.size();
+            // Using the timestamp of the latest pressure sample
+            final long timestamp = pressures.get(pressures.size() - 1).getTimestamp();
+            final Pressure pressure = new Pressure(timestamp, averagePressure);
```
hb0: See STAD-400

persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
```diff
@@ -699,24 +703,41 @@ public Double loadAscend(final long measurementIdentifier, final boolean forceGn
     Double ascendFromPressures(final List<TrackV6> tracks) {
         Double totalAscend = null;
         for (final TrackV6 track : tracks) {
-            Double ascend = null;
-            Double lastAltitude = null;
-            for (final Pressure pressure : track.getPressures()) {
 
+            // Calculate average pressure because some devices measure large pressure differences when
+            // the display-fingerprint is used and pressure is applied to the display: Pixel 6 [STAD-400]
+            // This filter did not affect ascend calculation of devices without the bug: Pixel 3a
```
hb0, line 711: See STAD-400

Without this fix, the ascend on the Pixel 6 would be
- 1-6 km instead of 90 m, while on the Pixel 3a it's 87 m.
- 200-600 m instead of 0-5 m, while on the Pixel 3a it's 0 m

With the two average filters applied, both return 87 m and ~ 0-5 m.

hb0 (commented):
Implemented PR feedback, fixed bugs in altitude calculation, added a filter to support devices like the Pixel 6 and refactored the models to fix the database schema.

muthenberg (approved):

persistence/src/main/java/de/cyface/persistence/PersistenceLayer.java
```diff
@@ -695,6 +854,81 @@ public List<Track> loadTracks(final long measurementIdentifier) throws CursorIsN
...
+                    continue;
+                }
+                Validate.notNull(pauseEventTime);
+                final long resumeEventTime = eventCursor
+                        .getLong(eventCursor.getColumnIndex(EventTable.COLUMN_TIMESTAMP));
+
+                // Collect all GeoLocationsV6 and Pressure points until the pause event
+                final TrackV6 track = collectNextSubTrackV6(locations, pressures, pauseEventTime);
+                // FIXME {@param locations} and {@param pressures} are expected to start with the first data point which
+                // is later in time than {@param pauseEventTime} or to the last data point if the earlier doesn't exist.
```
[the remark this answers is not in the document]
hb0: https://cyface.atlassian.net/browse/STAD-396

persistence/src/test/java/de/cyface/persistence/PersistenceLayerTest.java
```diff
@@ -0,0 +1,152 @@
...
+import androidx.test.platform.app.InstrumentationRegistry;
+
+import de.cyface.persistence.model.GeoLocationV6;
+import de.cyface.persistence.model.Pressure;
+import de.cyface.persistence.model.TrackV6;
+
+/**
+ * Tests the inner workings of the {@link PersistenceLayer}.
+ *
+ * FIXME: Add tests with multiple sub-tracks
```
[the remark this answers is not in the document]
hb0: https://cyface.atlassian.net/browse/STAD-396

hb0:
Tests werden ausgeführt in Side-CI, not the linked one.

https://app.bitrise.io/app/e15dba9d072a8bae

PR #262, opened 2023-02-13
Task/stad 428 add statistics functions from sdk6

hb0:
This PR adds the new statistics functionalities from SDK 6 to SDK 7.

The goal is to add the
- average speed and duration features like in the SDK 6 PR: https://github.com/cyface-de/android-backend/pull/257
- elevation feature, just in the same way and especially database structure like in the SDK 6 PR: https://github.com/cyface-de/android-backend/pull/258 

This makes sure, the database `v6` version `1` contains the same data schema.
- I compared the generated `persistence/schemas/DatabaseV6/1.json` and both are identical, which means the no errors were made porting the `v6` schema.

SDK 6 is still on `measures` database version `16` while SDK 7 is on `measures` database version `17` which I took into consideration while porting the changes from above. The only change between both versions is the `GeoLocationTable` which now contains `accuracy` as `double` in `meters` instead of `cm`.

When we migrate SDK 6 apps to SDK 7 the `measures` database version `16` will be automatically upgrade to `17` just like the SDK 7 apps did. The only migration step we need to add when it's time is to add a new Room database `v7` which loads and merges data from both previous databases, `measures.Version17` and `v6.Version1` into `v7.Version1`.


hb0 (commented):
pre-reviewed

persistence/build.gradle
```diff
@@ -54,9 +61,11 @@ android {
         targetCompatibility rootProject.ext.targetCompatibility
     }
 
-    // To avoid having to upgrade to `androidX.appCompat:1.3.0-beta1` [CY-5525]
     lintOptions {
+        // To avoid having to upgrade to `androidX.appCompat:1.3.0-beta1` [CY-5525]
         disable 'ObsoleteLintCustomCheck'
+        // Attention: Don't merge this into `main` (PR #257) <<<<<<<<<< FIXME
+        abortOnError false
```
hb0, line 68: FIXME

PR #268, opened 2023-02-23
[STAD-455] Use Singleton to create DatabaseV6

hb0:
This PR makes sure the `release-6` branch uses the `Room` database `v6` in a similar way we use the `Room` database on the `main` branch, so this method is tested in our test-app.

### DIFF
--- a/persistence/src/main/java/de/cyface/persistence/DatabaseV6.java
+++ b/persistence/src/main/java/de/cyface/persistence/DatabaseV6.java
@@ -18,7 +18,10 @@
  */
 package de.cyface.persistence;
 
+import android.content.Context;
+
 import androidx.room.Database;
+import androidx.room.Room;
 import androidx.room.RoomDatabase;
 
 import de.cyface.persistence.dao.GeoLocationDao;
@@ -42,12 +45,26 @@ import de.cyface.persistence.model.PersistedPressure;
  * where we cannot migrate to LiveData.
  *
  * @author Armin Schnabel
- * @version 1.0.0
+ * @version 1.1.0 // Port all future changes to `main` branch
  * @since 6.3.0
  */
 // (!) If `version` is increased, ensure migration to `measures` Version 18 is still possible.
-@Database(entities = {PersistedPressure.class, PersistedGeoLocation.class}, version = 1)
+@Database(
+        entities = {PersistedPressure.class, PersistedGeoLocation.class},
+        // If increased, ensure to add migration in `main` branch into `measures` Version 18
+        version = 1
+)
 public abstract class DatabaseV6 extends RoomDatabase {
+
+    /**
+     * The file name of the database represented by this class.
+     */
+    private static final String DATABASE_NAME = "v6";
+    /**
+     * Singleton to prevent multiple open database-instances at the same time.
+     * See https://developer.android.com/codelabs/android-room-with-a-view-kotlin#7
+     */
+    private static volatile DatabaseV6 instance;
 
     /**
      * @return Data access object which provides the API to interact with the {@link PersistedPressure} database table.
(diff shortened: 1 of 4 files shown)


persistence/src/main/java/de/cyface/persistence/DatabaseV6.java
```diff
@@ -42,21 +45,66 @@
...
+     * <p>
+     * Additional notes: Room itself (like SQLite, Room is thread-safe) and only uses one connection for writing.
+     * I.e. we only need to worry about deadlocks when running manual transactions (`db.beginTransaction` or
+     * `roomDb.runInTransaction`. See
+     * https://www.reddit.com/r/androiddev/comments/9s2m4x/comment/e8nklbg/?utm_source=share&utm_medium=web2x&context=3
+     * The PersistenceLayer constructor is called from main UI and non-UI threads, e.g.:
+     * - main UI thread: CyfaceDataCapturingService, DataCapturingBackgroundService/DataCapturingService,
+     * DataCapturingButton, MeasurementOverviewFragment
+     * - other threads: SyncAdapter, Event-/MeasurementDeleteController
+     */
```
hb0, line 97: That documentation moved here from `PersistenceLayer`

hb0 (commented):
pre-reviewed

muthenberg (changes requested):
Is there truely no alternative to using a Singleton, like instantiating exactly one instance of the database in your application class and providing that instance to all classes requiring it?

hb0:
I'm actually not even sure if using the `Singleton` ensured that only one instance of the `Database` is created, as we instantiate the `PersistenceLayer` (which holds the `database` instance) from different processes:
- `DataCapturingProcess` (created from foreground)
- `DataCapturingBackgroundProcess` (created in the background)
- `SyncAdapter`  (created by the system, but we could use `ContentProvider` instead)
- `DataCapturingButton/Fragements/Controllers` (UI)

or from some previous notes of mine:
```
The PersistenceLayer constructor is called from main UI and non-UI threads, e.g.:
* - main UI thread: CyfaceDataCapturingService, DataCapturingBackgroundService/DataCapturingService,
* DataCapturingButton, MeasurementOverviewFragment
* - other threads: SyncAdapter, Event-/MeasurementDeleteController
```

But the `Room` database usually always has to be access from an async thread (we add the async API in a later PR), so even if `SyncAdapter` uses `ContentProvider` and we keep the `ContentProvider` using the SQLite API like it's still the case, `Room` would be accessed from different threads.

I didn't want to do too much refactoring in this PR as the purpose of the PR was mostly to replace the SQLite API usage with the `Room` API, and for that to implement the migration code to merge the temporary `altitude` database into the main database.

The other question is, if it's really necessary to ensure that only one `Database` instance exists, or if this was just to simplify the Android sample code/guide.

------------

The Room documentation states:

```
From Room guide: https://developer.android.com/training/data-storage/room
* If the app runs in multiple processes, include enableMultiInstanceInvalidation() in the builder.
* That way, when when you have an instance of AppDatabase in each process, you can invalidate the shared
* database file in one process, and this invalidation automatically propagates to the instances of AppDatabase
* within other processes.
* <p>
* Additional notes: Room itself (like SQLite, Room is thread-safe) and only uses one connection for writing.
* I.e. we only need to worry about deadlocks when running manual transactions (`db.beginTransaction` or
* `roomDb.runInTransaction`. See
* https://www.reddit.com/r/androiddev/comments/9s2m4x/comment/e8nklbg/?utm_source=share&utm_medium=web2x&context=3
```

I'm not really sure what is meant by `you can invalidate the shared database file in one process` - when should we do so?
- https://androidx.tech/artifacts/room/room-runtime/2.1.0-alpha05-source/androidx/room/MultiInstanceInvalidationClient.java.html
=> I guess I just remove enabling this option because we don't use the "invalidate database file" function yet. Maybe this is to abort some long running modification operations, which are always executed in a transaction and, thus, lock the database?

From discussions I found, it seems like `Room` itself is thread-safe:
- https://github.com/android/architecture-components-samples/issues/60
- https://stackoverflow.com/questions/47525082/can-i-use-the-same-roomdatabase-object-from-multiple-threads-on-android

Before changing to the Room abstraction API we also didn't have to handle multi-process issues with SQLite and had no problems so far, even back then when we wrote the sensor data into the database.

=> I'll just remove the Singleton and remove `enableMultiInstanceInvalidation` with some documentation why we don't enable this option while using Room from multiple processes.

hb0:
As Room should be thread-safe and only one connection is used for write operations, there should not be the need to wrap the Database instantiation in a Singleton.

But we need `enableMultiInstanceInvalidation` as this allows us to use the same Room database from multiple processes without us having to implement additional stuff like "invalidate database file", see [here](https://commonsware.com/Room/pages/chap-processes-003.html):
- tells Room that we use it across processes
- Room sets up a `MultiInstanceInvalidationService` in the main process, instances in other processes connect to that service and db modification information is shared between the processes

PR #277, opened 2023-04-18
Task/stad 310 extract uploader

hb0:
see https://github.com/cyface-de/uploader/pull/1


hb0 (commented):
pre-reviewed

synchronization/src/main/java/de/cyface/synchronization/SyncAdapter.kt
```diff
@@ -177,17 +179,21 @@ class SyncAdapter private constructor(
                             for (listener in progressListener) {
                                 listener.onProgress(total.toFloat(), measurement.id)
                             }
-                        }, jwtAuthToken
+                        }
+                    }
+                    val result = syncPerformer.sendData(
+                        http, syncResult, endPointUrl, metaData,
+                        compressedTransferTempFile!!, processListener, jwtAuthToken
```
muthenberg: If there is a line break, please put each parameter on its own line. This will ease future reviewes and is the recommended style for Swift and Kotlin Code.

synchronization/src/cyface/java/de/cyface/synchronization/CyfaceAuthenticator.java
```diff
@@ -245,58 +234,18 @@ public Bundle hasFeatures(AccountAuthenticatorResponse response, Account account
...
-        // Generate login payload
-        final JSONObject loginPayload = new JSONObject();
-        try {
-            loginPayload.put("username", username);
-            loginPayload.put("password", password);
-        } catch (final JSONException e) {
-            throw new IllegalArgumentException(e);
-        }
+            NetworkUnavailableException, TooManyRequestsException, HostUnresolvable, ForbiddenException,
+            UnexpectedResponseCode, AccountNotActivated {
```
muthenberg, line 249: Would it be possible to wrap all these exceptions in a common one (like LoginFailed), reducing the footprint of this function, but still keeping the information? The Actual stack trace should still persist, wouldn't it?

muthenberg (changes requested):
See my comments inline

.github/workflows/gradle_publish.yml
```diff
@@ -17,26 +17,43 @@ jobs:
...
+          java-version: '17'
 
       - name: Add gradle.properties
         run: |
-          # Use the repository's automatically set up token to pushlish to the registry
+          # Use the repository's automatically set up token to publish to the registry
           cp gradle.properties.template gradle.properties
-          echo "githubUser=${{ github.actor }}" >> gradle.properties
-          echo "githubToken=${{ secrets.GITHUB_TOKEN }}" >> gradle.properties
+          ##### FIXME: using ci account while `uploader` is still private
```
hb0, line 31: TODO

.github/workflows/gradle_publish.yml
```diff
@@ -17,26 +17,43 @@ jobs:
...
+          echo "githubToken=${{ secrets.GH_READ_TOKEN }}" >> gradle.properties
+          # Disable jdk version setting to as it's defined above
+          sed -i "s/org.gradle.java.home=/#org.gradle.java.home=/g" gradle.properties
 
       # versionName is required to publish artifacts to Github Registry
       - name: Set versionName
         run: |
           sed -i "s/cyfaceAndroidBackendVersion = \"0.0.0\"/cyfaceAndroidBackendVersion = \"${{ github.ref_name }}\"/g" build.gradle
 
+      # FIXME: As we need the ci token above to download the uploader, we need to build here and then overwrite with the repo token to publish
```
hb0, line 42: TODO

.github/workflows/gradle_publish.yml
```diff
@@ -17,26 +17,43 @@ jobs:
...
+      - name: Use repo token to publish
+        run: |
+          # Use the repository's automatically set up token to publish to the registry
+          cp gradle.properties.template gradle.properties
+          echo "githubUser=${{ github.actor }}" >> gradle.properties
+          echo "githubToken=${{ secrets.GITHUB_TOKEN }}" >> gradle.properties
+          # Disable jdk version setting to as it's defined above
+          sed -i "s/org.gradle.java.home=/#org.gradle.java.home=/g" gradle.properties
+
+      # `build` added to fix cannot find symbol de.cyface.datacapturing.BuildConfig until upgraded to Gradle 8 [RFR-412]
```
hb0, line 54: TODO

build.gradle
```diff
@@ -45,9 +45,9 @@ ext {
     cyfaceAndroidBackendVersion = "0.0.0" // Automatically overwritten by CI
 
     // Cyface dependencies
-    cyfaceUtilsVersion = "3.3.5"
-    cyfaceSerializationVersion = "2.3.4"
-    cyfaceUploaderVersion = "1.0.0-alpha1" // FIXME
+    cyfaceUtilsVersion = "3.3.6"
+    cyfaceSerializationVersion = "2.3.7"
+    cyfaceUploaderVersion = "1.0.0-beta2" // FIXME
```
hb0, line 50: TODO

hb0 (commented):
pre-reviewed

build.gradle
```diff
@@ -45,8 +45,9 @@ ext {
     cyfaceAndroidBackendVersion = "0.0.0" // Automatically overwritten by CI
 
     // Cyface dependencies
-    cyfaceUtilsVersion = "3.3.5"
-    cyfaceSerializationVersion = "2.3.3"
+    cyfaceUtilsVersion = "3.3.7"
+    cyfaceSerializationVersion = "2.3.7"
+    cyfaceUploaderVersion = "1.0.0-beta6" // FIXME
```
hb0, line 50: FIXME

synchronization/src/cyface/java/de/cyface/synchronization/CyfaceAuthenticator.java
hb0: I've simplified the API of `Authenticator` and `Uploader` in `uploader` library to return `Login-/Registration-/UploadFailed` as exception, but in the SDK in `CyfaceAuthenticator` we still need some code to pass the exception type to the `ErrorHandler`:

```
} catch (final LoginFailed e) {
            var cause = e.getCause();
            if (cause instanceof ServerUnavailableException || cause instanceof ForbiddenException) {
                sendErrorIntent(context, SERVER_UNAVAILABLE.getCode(), e.getMessage());
                throw new NetworkErrorException(e);
            } else if (cause instanceof MalformedURLException) {
                sendErrorIntent(context, MALFORMED_URL.getCode(), e.getMessage());
                throw new NetworkErrorException(e);
            } else if (cause instanceof SynchronisationException) {
                sendErrorIntent(context, SYNCHRONIZATION_ERROR.getCode(), e.getMessage());
```

Maybe there is a better solution for error handling, but I don't want to refactor the SDK as the purpose of the PR is just to extract the auth/upload functions to the `uploader` library with a simple API for 3rd parties.

I think we're in a good spot now, with the current `uploader` interface:
```
public abstract fun upload(
  jwtToken: kotlin.String,
  metaData: de.cyface.model.RequestMetaData,
  file: java.io.File,
  progressListener: de.cyface.uploader.UploadProgressListener
): de.cyface.uploader.Result
// now only throws `UploadFailed` wrapper exception

```

.github/workflows/gradle_publish.yml
hb0: see if this works without the build step again

PR #300, opened 2024-05-18
Add logout support

hb0:
This PR adds code required for logout from the UI.

**Attention: There is still a bug in the logout which leads to a `Credentials incorrect`  error after logging in with another account. You need to restart the app to fix this. But this is still better than no logout functionality.**


hb0 (commented):
pre-reviewed

PR #304, opened 2024-09-26
[RFR-1170] Fix NPE during upload attempt

hb0:


synchronization/src/main/kotlin/de/cyface/synchronization/SyncAdapter.kt
```diff
@@ -615,8 +615,7 @@ class SyncAdapter private constructor(
             tracks.isEmpty() || (tracks[0].geoLocations.size > 0
                     && tracks[tracks.size - 1].geoLocations.size > 0)
         )
-        val lastTrack: List<ParcelableGeoLocation?>? =
-            if (tracks.isNotEmpty()) tracks[tracks.size - 1].geoLocations else null
+        val lastTrack = if (tracks.isNotEmpty()) tracks[tracks.size - 1].geoLocations else null
```
hb0, line 618: just formatting

synchronization/src/main/kotlin/de/cyface/synchronization/SyncAdapter.kt
```diff
@@ -641,7 +640,9 @@ class SyncAdapter private constructor(
             }
             val filesSize = if (allAttachments > 0) {
                 // TODO: support multiple attachments by zipping them
-                val attachment = persistence.attachmentDao!!.loadOneByMeasurementIdAndType(measurement.id, FileType.JSON)
+                // (!) we load the CSV file here as the location_metrics.csv should always exist
+                // if not, the upload crashes with an NPE.
+                val attachment = persistence.attachmentDao!!.loadOneByMeasurementIdAndType(measurement.id, FileType.CSV)
```
hb0, line 645: fix: we now always create a CSV file but the JSON file is optional, which was the cause for the NPE. This is a hot fix as we're planning to zip all meta attachment files anyway soon

hb0 (commented):
pre-reviewed

PR #310, opened 2025-02-12
[RFR-1213] Fix files size changes between attachment uploads

hb0:
I noticed that the `filesSize` in the metadata of the attachment uploads (object store upload) changes while uploading a measurement with lots of images.

The reason for this is that we used the folder size which changes during upload, as we delete uploaded images.

This PR (together with the other PR in this ticket) fixes this bug by introducing a `filesSize` column to the `Measurement` table which is updated when a new attachment is collected. We can then simply use this `filesSize` for the metadata, ensuring the metadata entry does not change randomly during upload. The PR also ensures we don't have to iterate over thousands of `attachment` table entries to calculate the `filesSize` before each upload but just calculate this number once (after each attachment collected).

I also tested the migration and the `filesSize` manually on a physical test device.


persistence/schemas/de.cyface.persistence.Database/20.json
hb0, line 1: This is auto-generated by Android Room.

persistence/src/main/kotlin/de/cyface/persistence/DatabaseMigrator.kt
```diff
@@ -92,6 +92,17 @@ class DatabaseMigrator(val context: Context) {
...
+         * Adds the [de.cyface.persistence.model.Measurement.filesSize] column.
+         */
+        val MIGRATION_19_20 = object : Migration(19, 20) {
+            override fun migrate(database: SupportSQLiteDatabase) {
+                // Add the new filesSize column with default value 0 for existing rows
+                // This should be fine as we don't upload attachments in production, yet.
+                database.execSQL("ALTER TABLE Measurement ADD COLUMN filesSize INTEGER NOT NULL DEFAULT 0")
+            }
+        }
+
```
hb0, line 105: As we don't use the object store attachments upload on production yet, we should be fine not don't have to calculate the files size for measurements collected with the previous version here but just use "0" instead.

synchronization/src/main/kotlin/de/cyface/synchronization/SyncAdapter.kt
```diff
@@ -71,6 +71,8 @@ import java.util.UUID
  *
  * @author Armin Schnabel
  * @author Klemens Muthmann
+ * @version 4.2.0
+ * @since 2.0.0
```
hb0, line 75: I checked the history for the last version and since used and added them again.

hb0 (commented):
pre-reviewed

